Issue #12, opened 2019-09-25
pav result -f should be single test only.


Paul-Ferrell:
It should probably pretty print the full result json.  A table doesn't work between tests with different result fields anyway.

Paul-Ferrell:
I should have made this ticket more clear. 
 1. Because tests will have different sets of results, the -f option doesn't make sense when requesting the results of more than one test. If you do so, an error should be printed. 
 2. The result table doesn't make sense for the -f option either. Use the pprint module to pretty print the json object instead.

PR #22, opened 2019-09-26
pav results -f only returns 1 test 

francinelapid:
#12


test/tests/extraneous_prints_tests.py
```diff
@@ -22,7 +22,7 @@ def test_for_extra_prints(self):
             "grep",
             "-R",
             "-I",
-            '[^f]print('
```
Paul-Ferrell: Look closer at this test. There's a way to note when pprint and print statements are acceptable via comments.

lib/pavilion/plugins/commands/result.py
```diff
@@ -91,12 +94,21 @@ def run(self, pav_cfg, args):
             title="Test Results"
         )
 
-    def _get_tests(self, pav_cfg, tests_arg):
-        if not tests_arg:
+    def _get_tests(self, pav_cfg, tests_arg, full_arg):
```
Paul-Ferrell: The logic of this really bugs me. How about this structure instead:
1. Get the tests based on the tests_arg (or absence thereof).
2. Throw an error if the full argument was given and there's more than one test, or print a warning that we've cut it down to just the first one.

Paul-Ferrell:
I made one little change, and approved.
 - The 'too many tests' warning goes to the user, not to log.

Issue #4, opened 2019-09-23
Tests need to be able to configure their timeout.


Paul-Ferrell:
- Add a new option under 'run' (and 'build') in the test config. 
 - The timeout is managed in the run method of the pav_test.PavTest class.
 - Accept null/None as no timeout.
 - Create a unittest test that times out. (after 1 second or something).

PR #25, opened 2019-09-30
Added 'timeout' option to build and run sections of the config allowi…

nicholas-sly:
…ng for user configuration.

Addresses #4


lib/pavilion/pav_test.py
```diff
@@ -487,14 +487,18 @@ def handle_error(_, path, exc_info):
...
+    NONE_LIST = ['', 'null', 'None', None]
 
     def _build(self, build_dir):
         """Perform the build. This assumes there actually is a build to perform.
         :param Path build_dir: The directory in which to perform the build.
         :returns: True or False, depending on whether the build appears to have
             been successful.
         """
+        if self.config['build']['timeout'] in self.NONE_LIST:
+            BUILD_SILENT_TIMEOUT = None
```
Paul-Ferrell: ALL_CAPS_NAMES are reserved for globals.

lib/pavilion/pav_test.py
```diff
@@ -515,7 +519,8 @@ def _build(self, build_dir):
                                         stdout=build_log,
                                         stderr=subprocess.STDOUT)
 
-                timeout = self.BUILD_SILENT_TIMEOUT
+                timeout = BUILD_SILENT_TIMEOUT if BUILD_SILENT_TIMEOUT is not \
```
Paul-Ferrell: 1. If you have to wrap it, it's too long for an inline if.
2. If there is not timeout, why set one? None is a valid argument for proc.wait.

lib/pavilion/pav_test.py
```diff
@@ -524,17 +529,20 @@ def _build(self, build_dir):
                         log_stat = build_log_path.stat()
                         quiet_time = time.time() - log_stat.st_mtime
                         # Has the output file changed recently?
-                        if self.BUILD_SILENT_TIMEOUT < quiet_time:
+                        if BUILD_SILENT_TIMEOUT is None:
```
Paul-Ferrell: This is unnecessary, if you just use `None` as the timeout.

lib/pavilion/pav_test.py
```diff
@@ -487,14 +487,18 @@ def handle_error(_, path, exc_info):
 
     # A process should produce some output at least once every this many
     # seconds.
-    BUILD_SILENT_TIMEOUT = 30
+    NONE_LIST = ['', 'null', 'None', None]
```
Paul-Ferrell: I'm putting this here so you can see it first. There are four cases.
 1. A valid user specified numerical timeout  - '3'
 2. An invalid user specified timeout (negative, non-int). - 'bob', '-1'
 3. No timeout. - null, 'None', etc
 4. The default timeout. - '' or not set.
 
First off, you don't check and give a good error message for case 2.
Secondly, just take an explicit `null` (which translates to None in python).

lib/pavilion/pav_test.py
```diff
@@ -756,6 +762,11 @@ def run(self, sched_vars, sys_vars):
         self.status.set(STATES.PREPPING_RUN,
                         "Resolving final run script.")
 
+        if self.config['run']['timeout'] in self.NONE_LIST:
```
Paul-Ferrell: Everything I said about the build timeout applies here too.

lib/pavilion/pav_test.py
nicholas-sly: I swear I tried to allow for it, but got an error.  I can try again.  It may have been something else.

lib/pavilion/pav_test.py
nicholas-sly: Check for invalid values is now in the init.

lib/pavilion/pav_test.py
```diff
@@ -178,6 +178,19 @@ def __init__(self, pav_cfg, config, sys_vars, _id=None):
...
+            if loc in config and 'timeout' in config[loc] and \
+                    config[loc]['timeout'] is not None:
+                try:
+                    test_timeout = int(config[loc]['timeout'])
+                    if test_timeout < 0:
+                        raise ValueError()
+                except ValueError:
+                    raise PavTestError("{} timeout must be a non-negative "
+                            "integer or empty.  Received {}.".format(
+                                loc, config[loc]['timeout']))
```
Paul-Ferrell: Completely normalize and convert the timeout here, and assign them to _build_timeout and _run_timeout. Then just use those later.

lib/pavilion/test_config/file_format.py
```diff
@@ -182,7 +182,13 @@ class TestConfigLoader(yc.YamlConfigLoader):
                               "compiled on the host that launches the test."),
                 yc.ListElem('cmds', sub_elem=yc.StrElem(),
                             help_text='The sequence of commands to run to '
-                                      'perform the build.')
+                                      'perform the build.'),
+                yc.StrElem(
+                    'timeout',
+                    default='30',
+                    help_text="Time that a build can continue without "
```
Paul-Ferrell: Time in what unit?

Paul-Ferrell:
Still waiting for you to address some changes.

Issue #17, opened 2019-09-25
Slurm tasks_per_node fails when 'all' is used.


Paul-Ferrell:

Paul-Ferrell:
This was already fixed.

PR #26, opened 2019-10-01
Added option for 'all' as 'tasks_per_node'.

nicholas-sly:
Addresses #17 .  Also adds 'True' and 'False' for available values to the immediate slurm option.


Paul-Ferrell:
This already got fixed elsewhere.

Issue #6, opened 2019-09-25
pav status should show some scheduling info. 


Paul-Ferrell:
This should be available via another method on scheduler plugins. A single text value should suffice.

Paul-Ferrell:
For the raw scheduler:
  - CPU and mem consumption?
For slurm:
  - Number of nodes and CPUs the job is running on.  Maybe time remaining.

Paul-Ferrell:
This information needs to be per-job, otherwise it doesn't make sense. You're going to have to query the scheduler about the particular job in question. For raw, that will mean asking the OS about the job somehow (/proc is a good place to look). 

For slurm, `scontrol show job <jobid>` will get you some data. There's already a method on the slurm scheduler for parsing scontrol output.

PR #34, opened 2019-10-15
Status with scheduler info

francinelapid:
#6 

I think I accidentally ignored merge conflicts for one or two of the files.

The last column of `pav status` output looks like this (for a test using the `raw` scheduler):
```
+--------------------------------------
|Scheduler Info
+--------------------------------------
| {'cpus:': '36', 'total_mem': '125798'} 
```

I tried it with the `slurm` scheduler, but the variables are kept as `Deferred`


Paul-Ferrell:
See ticket comments.

lib/pavilion/plugins/sched/raw.py
```diff
@@ -284,3 +285,24 @@ def _cancel_job(self, test):
...
+    def get_overall_status(self, test):
+
+        if not test.job_id:
+            return "Job ID not found."
+        else:
+            host, pid = test.job_id.rsplit('_', 1)
+
+        sched_stat = {}
+
+        regex = re.compile("(VmSize|Cpus_allowed_list):\s(.+)")
```
Paul-Ferrell: Cpus_allowed_list isn't super useful. See `man proc` for info on other useful info you could add from other files. 
`/proc/[pid]/sched` might be useful too. I'm leaving it up to you to do the research and figure out a reasonable set of data to pull.

lib/pavilion/plugins/sched/raw.py
```diff
@@ -284,3 +285,24 @@ def _cancel_job(self, test):
                 STATES.SCHED_ERROR,
                 "PID {} refused to die.".format(pid)
             )
+
+    def get_overall_status(self, test):
+
+        if not test.job_id:
```
Paul-Ferrell: If a test doesn't have a job id, then it was never run.

lib/pavilion/plugins/sched/raw.py
francinelapid: Instead of `Cpus_allowed_list` I used `se.sum_exec_runtime` from the `sched` file

lib/pavilion/plugins/commands/status.py
```diff
@@ -36,6 +41,7 @@ def status_from_test_obj(pav_cfg, test_obj):
             'state': status_f.state,
             'time': status_f.when.strftime("%d %b %Y %H:%M:%S %Z"),
             'note': status_f.note,
+            'scheduler info': scheduler_status
```
Paul-Ferrell: This is not ok. 

You need to read the documentation on this object more. Output is limited to the size of a block to ensure the file append is atomic. 

Adding to this object is a pretty major change, and do we need to know this information at every stage?

This makes me think that maybe this whole concept should be folded into the general scheduler status, and then incorporated into the status note in those situations.

lib/pavilion/plugins/sched/raw.py
Paul-Ferrell: The check is fine, the message is still not correct.

lib/pavilion/plugins/sched/raw.py
```diff
@@ -284,3 +285,34 @@ def _cancel_job(self, test):
...
+            return "Job ID not found."
+        else:
+            host, pid = test.job_id.rsplit('_', 1)
+
+        sched_stat = {}
+
```
Paul-Ferrell: I think this looks great.

Paul-Ferrell:
I don't think the status object is the place to store it. That's for tracking a history, not keeping information like this. I think this is something the _run command should trigger through the scheduler (given a PavTest object), and saved using the PavTest object to another file in the test's run directory, much like jobid is handled now. Since we're making this standard, we should always include the data in the results object.

lib/pavilion/plugins/commands/status.py
```diff
@@ -29,12 +29,15 @@ def status_from_test_obj(pav_cfg, test_obj):
             sched = schedulers.get_plugin(test.scheduler)
             status_f = sched.job_status(pav_cfg, test)
 
+        if status_f.state in [STATES.SCHEDULED, STATES.RUNNING]:
+            status_f.note += " " + str(test.sched_info)
```
francinelapid: not sure what other states I should add

PR #35, opened 2019-10-21
Fixed num_procs issues.

Paul-Ferrell:


lib/pavilion/plugins/sched/slurm.py
```diff
@@ -38,8 +39,11 @@ def get_lines(self):
             lines.append('#SBATCH --account {s._conf[account]}'.format(s=self))
 
         lines.append('#SBATCH -N {s._nodes}'.format(s=self))
-        lines.append('#SBATCH --tasks-per-node={s._conf['
-                     'tasks_per_node]}'.format(s=self))
+        tasks = self._conf['tasks_per_node']
+        if tasks == 'all':
+            tasks = self._vars['min_ppn']
+        lines.append('#SBATCH --tasks-per-node={tasks_per_node}'
+                     .format(tasks_per_node=tasks))
```
nicholas-sly: Is it necessary to set a variable just for use in the line append statement?  Can you not just put 'tasks', since it was defined on the previous two lines?

Paul-Ferrell: Yes, this is silly.

PR #38, opened 2019-10-22
Slurm mpi sched

francinelapid:
Very similar to the Slurm scheduler plugin, only it uses `mpirun` and not `srun`


lib/pavilion/plugins/sched/slurm_mpi.py
```diff
@@ -0,0 +1,26 @@
+from pavilion.schedulers import SchedulerPlugin
+from pavilion.schedulers import SchedulerPluginError
+from pavilion.schedulers import SchedulerVariables
+from pavilion.schedulers import dfr_var_method
+import pavilion.plugins.sched.slurm as slurm
+
+
+class SlurmMPIVars(slurm.SlurmVars):
+    @dfr_var_method
+    def test_cmd(self):
```
Paul-Ferrell: This has to build a command up to like the original. It needs to specify the actual number of requested nodes for a test, and the number of procs to use. 

Note that the mpirun defaults are not the same as those for srun.

francinelapid: I couldn't find an `mpirun` argument that specifies the number of nodes, just the number of processes per node.

Paul-Ferrell: That will need to be specified then. Luckily, it's how we already tell the slurm plugin to handle it anyway.

francinelapid: something like this works:
```
scheduler: slurm_mpi
slurm_mpi:
    num_nodes: 2
```

PR #39, opened 2019-10-22
General Table Result Parser

francinelapid:


lib/pavilion/plugins/results/table.py
```diff
@@ -0,0 +1,98 @@
...
+            name='table',
+            description="Parses tables"
+        )
+
+    def get_config_items(self):
+
+        config_items = super().get_config_items()
+        config_items.extend([
+            yc.StrElem(
+                'delimiter', required=True,
```
Paul-Ferrell: This should default to whitespace.

lib/pavilion/plugins/results/table.py
```diff
@@ -0,0 +1,98 @@
...
+                help_text="Constant that will be placed in result."
+            ),
+            yc.StrElem(
+                'col_num', required=True,
```
Paul-Ferrell: Should default to the number of column names if those are given.

lib/pavilion/plugins/results/table.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+    def __call__(self, test, file, delimiter=None, col_num=None,
+                 row_col='', col_names = []):
+
+        match_list = []
+
+        # generate regular expression
+        value_regex = '(\S+| )'
+        new_delimiter = ' *' + delimiter + ' *'
```
Paul-Ferrell: It's probably better to use '\s*' rather than ' *' here.

Paul-Ferrell: And in general.

lib/pavilion/plugins/results/table.py
```diff
@@ -0,0 +1,98 @@
...
+                'delimiter', required=True,
+                help_text="Constant that will be placed in result."
+            ),
+            yc.StrElem(
+                'col_num', required=True,
+                help_text="Number of columns in table, including row names, "
+                          "if there is such a column."
+            ),
+            yc.StrElem(
+                'row_col', default='False',
```
Paul-Ferrell: This is a confusing name. Maybe 'has_header'?

lib/pavilion/plugins/results/table.py
```diff
@@ -0,0 +1,98 @@
...
+        new_delimiter = ' *' + delimiter + ' *'
+        value_regex_list = []
+        for i in range(int(col_num)):
+            value_regex_list.append(value_regex)
+        str_regex = new_delimiter.join(value_regex_list)
+        str_regex = '^ *' + str_regex + ' *$'
+
+        regex = re.compile(str_regex)
+        for line in file.readlines():
+            match_list.extend(regex.findall(line))
```
Paul-Ferrell: `find_all` isn't really meant for this circumstance. You should just use regex.match.

lib/pavilion/plugins/results/table.py
francinelapid: I check this logic in `_check_args`

lib/pavilion/plugins/results/table.py
francinelapid: If there is a column for row names, the regex won't necessarily match the entire line.

Paul-Ferrell: You're right, I didn't read that code carefully enough.

PR #40, opened 2019-10-22
Scheduler plugin documentation

francinelapid:
Documentation on how to write scheduler plugins. Not sure if this is sufficient


Paul-Ferrell:
I'm going to take this bit over. It's a bit much to ask of someone who hasn't written a scheduler plugin to write the documentation on how to do it.

What you have is fine, it's just very incomplete.

Issue #42, opened 2019-10-22
Relative timestamp format


Paul-Ferrell:
Write a function (in the utils.py library) that takes a timezone aware datetime object and produces a time string with information relative to the current time.

Only include each of the following if they differ from a regular timestamp. 
  - timezone (based on the UTC offset, not the timezone name). 
  - year
  - day/month

The easiest way to do this is to dynamically build the datetime.strftime format string.

Paul-Ferrell:
Oh, and use this new format in all command printouts. You can do this through the draw_table options (ask Paul).

j-ogas:
TODO: 
- [ ] Remove all timezone stuff.

j-ogas:
Closing via #49.

PR #49, opened 2019-10-28
add relative timestamp function

j-ogas:
Addresses #42.

Note, I alphabetized the import statements (unrelated to the issue at hand but it bothered me)


lib/pavilion/utils.py
```diff
@@ -1,16 +1,38 @@
...
+import shutil
 import subprocess
 import sys
 import textwrap
-import shutil
-import itertools
 from collections import defaultdict
+from pathlib import Path
+
+def get_relative_timestamp(obj):
```
Paul-Ferrell: `base_dt`

j-ogas:
@pflarr Let me know when I can resume working on this. Last we chatted, we needed pav to pass an object (instead of string) for the transformation in `draw_tables` to work correctly.

lib/pavilion/utils.py
```diff
@@ -1,16 +1,39 @@
...
 import sys
 import textwrap
-import shutil
-import itertools
 from collections import defaultdict
+from pathlib import Path
+
+def get_relative_timestamp(base_dt):
+    """Print formatted time string based on the delta of time objects.
+    :param obj datetime object
```
Paul-Ferrell: :param <type> <name>: <description>

:param datetime base_dt: The timestamp to format.

lib/pavilion/utils.py
```diff
@@ -1,16 +1,39 @@
...
+    :rtype string
```
Paul-Ferrell: :rtype: str

lib/pavilion/utils.py
```diff
@@ -1,16 +1,39 @@
...
+    """
+    now = datetime.datetime.now()
+    format_ = ['%Y', '%b', '%a', '%H:%M:%S'] # year, month, day, time
+
```
Paul-Ferrell: for i in range(len(format_)):
  if now.strftime(format_[i]) != ....:
     return now.strftime(" ".join(format_[i:])

lib/pavilion/utils.py
```diff
@@ -222,8 +238,8 @@ class PavEncoder(json.JSONEncoder):
     def default(self, o):  # pylint: disable=E0202
         if isinstance(o, Path):
             return super().default(str(o))
```
Paul-Ferrell: This is a bug (in my code, not yours). It should just `return str(o)` here.

lib/pavilion/utils.py
```diff
@@ -222,8 +238,8 @@ class PavEncoder(json.JSONEncoder):
...
-        else:
-            return super().default(str(o))
+        if isinstance(o, (datetime.datetime)):
+            return o.isoformat()
```
Paul-Ferrell: and `return super().default(o)` otherwise.

Issue #28, opened 2019-10-03
Raise error if source_download_name is set without a source_location.


Paul-Ferrell:
It's easy and confusing to set the wrong value here in the build section.

PR #55, opened 2019-11-14
Raise error if source_download_name is without source_location

francinelapid:
#28


francinelapid:
I made a test for this but I'm not sure if it's necessary.

lib/pavilion/plugins/commands/run.py
```diff
@@ -148,6 +149,11 @@ def run(self, pav_cfg, args, out_file=sys.stdout, err_file=sys.stderr):
         failed_build = None
         # Building any tests that specify that they should be built before
         for test in all_tests:
+            # make sure source_download_name is not set without source_location
```
Paul-Ferrell: 1. You're raising an exception of a type that won't be caught, which is a bug.
2. If the exception were caught and printed as an error, the user would have no idea where the problem is.
3. The PavTest object is responsible for making sure the test config is valid, so this code belongs there.

Paul-Ferrell:
I don't think you need a test either.

lib/pavilion/test_run.py
```diff
@@ -436,6 +437,16 @@ def build(self):
 :return: True if these steps completed successfully.
 """
 
+        if self.config['build']['source_download_name'] is not None:
```
Paul-Ferrell: Move this to __init__, and raise an error. That way we can immediately see that there's a problem.

PR #61, opened 2019-11-26
Verify module loaded

francinelapid:
#9 

`verify_module_loaded` doesn't require a test_id anymore, but now we can't set the status...

I left the original lines in as comments for comparison.


Paul-Ferrell:
This was already fixed in another merge request.

Issue #46, opened 2019-10-23
Add `pav file[s]` command.


Paul-Ferrell:
`pav file <jobid>` should list all the files in a test run's run directory. 

`pav file <jobid> filename` should output that files contents.

j-ogas:
Closing via #63.

PR #63, opened 2019-11-26
add job id cat, ls, and tree commands

j-ogas:
Work in progress.

Addresses #46.

TODO:
- [x] determine what to do (if anything) about sub directories and files. For example, `pav_cfg.working_dir/'tests'/job_id/` has a `build` directory with files. Do we care?
- [x] add try-catch block around reading, i.e., guard against directory `--filename` arguments
- [ ] add travis tests
- [ ] make travis happy

### DIFF
--- a/lib/pavilion/plugins/commands/file.py
+++ b/lib/pavilion/plugins/commands/file.py
@@ -30,16 +30,18 @@ class FileCommand(commands.Command):
         test_dir = pav_cfg.working_dir/'tests'
         job_dir = test_dir/args.job_id
 
-        if os.path.isdir(job_dir.as_posix()) is True:
-            if args.filename is not None:
-                print_file(job_dir/args.filename)
-            else:
-                for file_ in os.listdir(job_dir):
-                    utils.fprint(file_, file=sys.stdout)
-        else:
-            utils.fprint("file '{}' does not exist.".format(job_dir.as_posix()),
+        if os.path.isdir(job_dir.as_posix()) is False:
+            utils.fprint("directory '{}' does not exist."
+                         .format(job_dir.as_posix()),
                          file=sys.stderr, color=utils.RED)
             sys.exit()
+
+        if args.filename is not None:
+            print_file(job_dir/args.filename)
+        else:
+            for file_ in os.listdir(job_dir):
+                utils.fprint(file_, file=sys.stdout)
+
         return 0
 
 
--- a/lib/pavilion/plugins/commands/file.py
+++ b/lib/pavilion/plugins/commands/file.py
@@ -35,27 +35,29 @@ class FileCommand(commands.Command):
                 print_file(job_dir/args.filename)
             else:
                 for file_ in os.listdir(job_dir):
-                    utils.fprint(file_, file=out_file)
+                    utils.fprint(file_, file=sys.stdout)
         else:
             utils.fprint("file '{}' does not exist.".format(job_dir.as_posix()),
-                         file=err_file)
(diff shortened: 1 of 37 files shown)


lib/pavilion/plugins/commands/file.py
```diff
@@ -0,0 +1,65 @@
...
+import sys
+
+from pavilion import commands
+from pavilion import utils
+
+class FileCommand(commands.Command):
+
+    def __init__(self):
+        super().__init__(
+            'file',
```
Paul-Ferrell: Let's call this the 'cat' command, and have an 'ls' and 'file' alias for it.

lib/pavilion/plugins/commands/file.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+        parser.add_argument(
+            '--filename',
+            type=str,
+            help="name of a file within working_dir/tests/<job id>/"
+        )
+
+    def run(self, pav_cfg, args):
+
+        test_dir = pav_cfg.working_dir/'tests'
```
Paul-Ferrell: The name of the test directory changed to `pav_cfg.working_dir/'test_runs'`

lib/pavilion/plugins/commands/file.py
```diff
@@ -0,0 +1,65 @@
...
+        else:
+            for file_ in os.listdir(job_dir):
+                utils.fprint(file_, file=sys.stdout)
+
+        return 0
+
+
+def print_file(filename):
+    try:
+        with open(filename, 'r') as file_:
```
Paul-Ferrell: `file` isn't a _builtin_ anymore. You can use it as a variable name.

lib/pavilion/plugins/commands/file.py
```diff
@@ -0,0 +1,65 @@
...
+            while True:
```
Paul-Ferrell: This doesn't matter here, but starting in python3.8:
```python
while block := file.read(4096):
  utils.fprint(block, file=sys.stdout)
```

Paul-Ferrell:
Check out the the tree command. 
We don't have to do anything that fancy, but just a basic tree would be cool.

Paul-Ferrell:
And it would be really nice to have a way to get metadata, maybe from a `-l` option?

lib/pavilion/plugins/commands/file.py
j-ogas: I wonder if this should just be split into two separate plugin commands, e.g., `commands/cat.py` and `commands/ls.py`?

j-ogas:
example of `pav ls`:
```bash
$ ./bin/pav ls 0000229
Could not find a pavilion config file. Using an empty/default config.
config
status
build.sh
build_hash
run.tmpl
build/
  stream.c -> /home/jogas/git/pavilion2/stream.c
  pav_build_log -> /home/jogas/git/pavilion2/pav_build_log
  stream -> /home/jogas/git/pavilion2/stream
  .built_by -> /home/jogas/git/pavilion2/.built_by
jobid
slurm.log
kickoff.log
```

One thing we need to discuss about `pav cat` is how to pass a filename. Using the example above, what would be the correct way to specify a file that belongs to a non-top level directory? For example: `0000229/build/pav_build_log`. My thoughts are:

1) `pav cat 0000229 build/pav_build_log`; or
2) `pav cat 0000229 pav_build_log`

Option one is much easier to implement. Option two is cleaner but introduces a corner case where there may be files with the same name present in different level directories. What are your thoughts @pflarr ?

PR #64, opened 2019-11-26
Exception tracking (and other config stuff).

Paul-Ferrell:
- Exceptions are now logged to a splunk compatible log file with extra debug info.
  - The location is configurable via the `exception_log` pavilion config attribute. 
- The current directory is no longer used as a config directory.
- Using the user's .pavilion directory can now be disabled via the `user_config`  pavilion config attribute.
- The pav source directory is now always used as a plugin directory.


bin/pav.py
```diff
@@ -99,6 +100,21 @@ def record_factory(*fargs, **kwargs):
     result_logger.setLevel(logging.INFO)
     result_logger.addHandler(result_handler)
 
+    # Setup the exception logger.
+    # Exceptions will be logged to this directory, along with other useful info.
+    exc_logger = logging.getLogger('exceptions')
+    exc_handler = RotatingFileHandler(
+        filename=pav_cfg.exception_log.as_posix(),
+        maxBytes=20 * 1024 **2,
```
j-ogas: What is the significance of 20 MB?

Paul-Ferrell: It's a random number.

Paul-Ferrell: Ok, I guess it's an arbitrary number. I'm really balancing overall file size with not losing useful log information if Pavilion ever puked a crazy amount of log data.

Issue #70, opened 2019-12-04
set status issue in pav-lib.bash.


Paul-Ferrell:
In pav-lib.bash
            ${PAV_PATH} set_status -s ENV_FAILED -n "${msg}" ${test_id}

On cray, the message argument must be given without a space between the -n and the message.

Issue #67, opened 2019-12-02
format table bug


CalvinDSeamons:
Currently If the size of the the terminal is too small it will delete information rather than wrapping it. I'm not sure if this is the intended functionality but I feel it should not occur.  
```console
[$user@$machine pavilion2]$ pav status
 Test statuses                                                                                                                       
---------+----------------+----------+----------+--------------------------------------------------------------------------------------
 Test id | Name           | State    | Time     | Note                                                                                 
---------+----------------+----------+----------+--------------------------------------------------------------------------------------
 1       | something5.two | SKIPPED  | 09:10:23 | Condition invalid. Not if user is calvindseamons. The current user is calvindseamons 
 2       | something5.one | COMPLETE | 09:10:23 | The test completed with result: PASS                                                 

[$user@$machine pavilion2]$ pav status
 Test statuses                                                                             
---------+----------------+----------+----------+--------------------------------------------
 Test id | Name           | State    | Time     | Note                                       
---------+----------------+----------+----------+--------------------------------------------
 1       | something5.two | SKIPPED  | 09:12:47 | Condition invalid. Not if user is          
         |                |          |          | calvindseamons. The current user is        
 2       | something5.one | COMPLETE | 09:12:47 | The test completed with result: PASS       

```
Note. `The current user is ` cuts off calvindseamons in the second pav status id 1.  
To confirm it wasn't anything with my skipped functions it also occurs on `COMPLETED` tests.
```console
[$user@$machine pavilion2]$ pav status
 Test statuses                                                                           
---------+--------------------+----------+----------+--------------------------------------
 Test id | Name               | State    | Time     | Note                                 
---------+--------------------+----------+----------+--------------------------------------
 1       | something.raw_test | COMPLETE | 09:17:56 | The test completed with result: PASS 
 2       | something.raw_test | COMPLETE | 09:17:56 | The test completed with result: PASS 

[$user@$machine pavilion2]$ pav status
 Test statuses                                              
---------+-------------+----------+----------+----------------
 Test id | Name        | State    | Time     | Note           
---------+-------------+----------+----------+----------------
 1       | something.r | COMPLETE | 09:18:03 | The test       
         |             |          |          | completed with 
 2       | something.r | COMPLETE | 09:18:03 | The test       
         |             |          |          | completed with 

```
Note. The second `pav status` chops off the `PASS` was well as `.raw_test`  

[...]

PR #79, opened 2019-12-11
Variable storage

Paul-Ferrell:
- Deferred are more flexible
   - They can contain multiple values
   - They don't need to be named, or have their elements pre-defined.
   - Variables that reference other variables become deferred if
     they reference something deferred.
 - Each test run is handed its variable manager now
   - The variables are saved and loaded when a run object is loaded.
   - Provided mechanisms for updating the remaining deferred vars.
   - Provided mechanisms for updating the test run config too.
   - Added a finalize method to do the two prior things in one shot.
 - The run.tmpl script isn't really a template any more, just a trial run.
   - run.sh is now written completely from scratch after deferred
     variables are resolved.
 - We can now add any pavilion variable value directly to
   the result defaults.
   - Added sys_name, user, and job_id (which isn't a var, but still).
- Output functions (fprint, etc) are now in output.py.

Compatability Breaking:
 - Changed the argument for SystemVariables from plugin_name to name


Paul-Ferrell:
Fixes #70

Paul-Ferrell:
Fixes #67

test/tests/test_setup_tests.py
```diff
@@ -287,6 +288,43 @@ def test_for_circular_variable_references(self):
...
+        fin_var_man.add_var_set('sys', undefered_sys_vars)
+
+        test.finalize(fin_var_man)
+
+        results = test.gather_results(test.run())
+        test.save_results(results)
+
+        plugins._reset_plugins()
+
     def test_resolve_all_vars(self):
```
nicholas-sly: Consider updating the test name to match the new function name (resolve_config).

test/tests/wait_cmd_tests.py
```diff
@@ -1,9 +1,7 @@
 from pavilion import commands
 from pavilion import plugins
-from pavilion import status_file
-from pavilion import system_variables
 from pavilion.series import TestSeries
-from pavilion.test_config import file_format
+from pavilion.test_config import file_format, VariableSetManager
 from pavilion.unittest import PavTestCase
 from pavilion.test_run import TestRun
```
nicholas-sly: Bump over unittest.

bin/pav-lib.bash
```diff
@@ -316,20 +316,20 @@ function verify_module_loaded() {
         1)
             msg="Module ${module_name}, version ${module_version} was not "
             msg="${msg}loaded. See the test log."
-            ${PAV_PATH} set_status -s ENV_FAILED -n "${msg}" ${test_id}
+            ${PAV_PATH} set_status -s ENV_FAILED -n"${msg}" ${test_id}
```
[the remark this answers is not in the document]
Paul-Ferrell: There's a ticket on the subject. On cray, for some reason, this doesn't work with the space.

PR #99, opened 2019-12-19
Cleaned up test completion states

Paul-Ferrell:
- Removed RUN_FAILED as a state.
   - Runs now return True (if the run.sh returned zero), false otherwise.
   - They may also raise a TimeoutError.
 - Made the create of the 'RUN_COMPLETE' file happen anytime a test run
   encounters an error that will prevent it from continuing.


Paul-Ferrell:
Fixes #98

Issue #86, opened 2019-12-12
bug: pav view <job id>


j-ogas:
Raises an exception.

Paul-Ferrell:
Should issue a warning and exit.

Paul-Ferrell:
This is still an issue.

CalvinDSeamons:
Yeah it `was` resolved. But not any more it would appear.

PR #112, opened 2020-01-08
Pavview

CalvinDSeamons:


CalvinDSeamons:
#86

lib/pavilion/plugins/commands/view.py
```diff
@@ -64,20 +64,23 @@ def run(self, pav_cfg, args):
...
+                host=args.host,
+                test_files=[],
+                tests=tests,
+                modes=args.modes,
+                overrides=overrides,
+                sys_vars=sys_vars,
+            )
+        except Exception as err:
+            fprint(err, file=self.errfile, color=31)
+            quit(0)
```
francinelapid: `sys.exit(<something other than 0*>)`

* 0 denotes a "successful termination"

lib/pavilion/plugins/commands/view.py
```diff
@@ -68,16 +68,19 @@ def run(self, pav_cfg, args):
...
+                host=args.host,
+                test_files=[],
+                tests=tests,
+                modes=args.modes,
+                overrides=overrides,
+                sys_vars=sys_vars,
+            )
+        except Exception as err:  # pylint: disable=W0703
+            fprint(err, file=self.errfile, color=31)
+            sys.exit(1)
```
Paul-Ferrell: - Return an error code, don't exit (the base pavilion script will handle exiting properly). 
 - Print a useful error, not just the exception.
 - Use the utils.COLOR constants, not just numbers. 
 - Figure out what exceptions could be raised, and catch those. Don't catch generic exceptions.

Issue #89, opened 2019-12-12
New Result Parser: Check for file existence


nicholas-sly:
Add a result parser that accepts a file name and returns whether or not it found that file in the working directory.

PR #117, opened 2020-01-15
add result parser to check if file exists in test_runs working_dir

CalvinDSeamons:
addresses issue #89


lib/pavilion/plugins/results/filecheck.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class Filecheck(result_parsers.ResultParser):
+    """Checks the working directory for the given file name.
+    The parser will tell the user if the filename exists or not. """
+
+    def __init__(self):
+        super().__init__(name='filecheck',
+                         description="Checks working directory"
+                         "for given file name")
```
nicholas-sly: Checks for the file, not the file name.  Same for the docstring above.

lib/pavilion/plugins/results/filecheck.py
```diff
@@ -0,0 +1,51 @@
...
+                                 "takes in a string and checks"
+                                 "the working directory for said "
+                                 "filename. It will return True or "
+                                 "False.")
+        ])
+
+        return config_items
+
+    def _check_args(self, filename=None):
+        if filename == "" or filename is None:
```
nicholas-sly: In python, the empty string "" and None evaluate to false, so you can just check `if not filename:`.

lib/pavilion/plugins/results/filecheck.py
```diff
@@ -0,0 +1,51 @@
...
+            raise result_parser.ResultParserError(
+                "File name cannot be null"
+            )
+
+    def __call__(self, test, file, filename):
+        try:
+            pav_cfg = config.find()
+        except Exception as err:
+            sys.exit(-1)
```
nicholas-sly: Paul is going to ask you to change this to the ERRNO stuff.  Ask Jordan.

Paul-Ferrell:
Most of your problems are because the guide for writing this type of plugin is incomplete. Feel free to bug me if you're unsure about these things in the future, and it will remind me to write a better guide.

lib/pavilion/plugins/results/filecheck.py
Paul-Ferrell: Nope, even worse. 

You shouldn't be using sys.exit at all. Let the exception happen, Pavilion will catch it, and error report it.

Paul-Ferrell: In some situations you do want to catch exceptions and re-raise them with a more meaningful message (that also includes the original message), usually along the lines of:
```
try: 
    stuff...
except (ExpectedException, OtherException) as err:
    raise ModuleSpecificException(
        "This specific bad thing happened in this specific place related to test '{test.id}': {err}"
        .format(test=test, err=err)

You should only catch those specific exceptions you expect, and never the really generic ones like Exception or BaseException.

File: lib/pavilion/plugins/results/filecheck.py [hunk 4]
```diff
@@ -0,0 +1,56 @@
...
+
+    def __init__(self):
+        super().__init__(name='filecheck',
+                         description="Checks working directory"
+                         "for a given file")
+
+    def get_config_items(self):
+        # Result parser consists of 1 string elem: filename.
+        config_items = super().get_config_items()
+        config_items.extend([
```
Paul-Ferrell: You don't need this argument. There is already a generic 'files' argument that the result parser uses to find (and open for you) specific files. See https://github.com/hpc/pavilion2/blob/6545b10d5f084ba0f8c996000f9896e39361c74b/lib/pavilion/result_parsers.py#L193

The files provide via the 'files' argument are found and opened for you by default, but you can change that behavior for your result parser plugin with arguments to __init__ (https://pavilion2.readthedocs.io/en/latest/source/result_parsers.html#pavilion.result_parsers.ResultParser.__init__)

lib/pavilion/plugins/results/filecheck.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def _check_args(self, filename=None):
+        # Filename can being anything but none.
+        if not filename:
+            raise result_parser.ResultParserError(
+                "File name cannot be null"
+            )
+
+    def __call__(self, test, file, filename):
+        # try to create config to get info on job.
```
Paul-Ferrell: You get passed the test object itself. So most of this path wrangling is unnecessary. See https://pavilion2.readthedocs.io/en/latest/source/test_run.html

`test.path` is what you're looking for.

lib/pavilion/plugins/results/filecheck.py
```diff
@@ -0,0 +1,56 @@
...
+        try:
+            pav_cfg = config.find()
+        except Exception as err:
+            return errno.EEXIST
+
+        # Build the test path of the job using make_id_path.
+        working_dir = pav_cfg.working_dir/'test_runs'
+        test_path = utils.make_id_path(working_dir, test.id)
+        # recursively search folders in path for filename.
+        for file in Path(str(test_path)).rglob(filename):
```
Paul-Ferrell: This should return true or false.

lib/pavilion/plugins/results/filecheck.py
```diff
@@ -0,0 +1,56 @@
...
+                                 "the working directory for said "
+                                 "file. It will return True with "
+                                 " the path-to-file or False.")
+        ])
+
+        return config_items
+
+    def _check_args(self, filename=None):
+        # Filename can being anything but none.
+        if not filename:
```
francinelapid: you don't to check for  this if you indicated that this config element is required, which you did.

Source: [Config Element class in yaml_config](https://github.com/lanl-preteam/yaml_config/blob/92424cdf2d5f89f477d0fe74627ef9d4dcda6e30/lib/yaml_config/elements.py#L113)

CalvinDSeamons:
Let me preface this by saying I'm probably missing something here, but this is what I've found. 
I understand using supers `file` as it is redundant to add another file element. However, by using `file` the tests are resolved in `run_cmd` when `parse_result` is called, a function that cannot be overwritten as it sits outside `result_parser.py`. When `file` is given a invalid filename (doesn't exist in the working dir) it is caught as an OSERROR and ended. This result in my own plugins `__call__` never being referenced thus the result is still `None` not Fail or False as desired in the ticket. 

You mentioned I can change the behavior of result_parser with arguments to `init` but without being able to override `parse_results` I don't see an easy way to reference the test when a file is not found.

Issue #132, opened 2020-02-03
pav wait should have an option to not timeout


francinelapid:
The default timeout is 60 seconds. The default should either be infinite timeout or at least have that as an option

```
optional arguments:
  -h, --help            show this help message and exit
  -j, --json            Give output as json, rather than as standard human
                        readable.
  -t TIMEOUT, --timeout TIMEOUT
                        Maximum time to wait for results in seconds.
                        Default=60.
```

PR #133, opened 2020-02-03
pav wait --no-timeout flag

francinelapid:
Added an option to `pav wait` to not have a timeout #132 

- [x] default should be no timeout (remove `--no-timeout` flag)
- [x] `--silent` (prints every 5 seconds)
- [x] change conditional
- [x] make status output prettier

Note:
- If using the `timeout` flag (`pav wait --timeout <value>`) a value is now required.
- Updates look something like this: 
`Wed Mar  4 16:01:24 2020 : test # 41 timeout.basic RUNNING Currently running.`

### DIFF
--- a/lib/pavilion/plugins/commands/wait.py
+++ b/lib/pavilion/plugins/commands/wait.py
@@ -74,12 +74,12 @@ class WaitCommand(commands.Command):
             if not args.silent:
                 if time.time() > (start_time + 5*periodic_status_count):
                     for test in tmp_statuses:
-                         stat = [str(time.ctime(time.time())), ':',
-                                 str(test['test_id']),
-                                 test['name'],
-                                 test['state'],
-                                 test['note']]
-                         fprint(' '.join(stat))
+                        stat = [str(time.ctime(time.time())), ':',
+                                  str(test['test_id']),
+                                  test['name'],
+                                  test['state'],
+                                  test['note']]
+                        fprint(' '.join(stat))
                     periodic_status_count += 1
 
         ret_val = status.print_status(tmp_statuses, self.outfile, args.json)
--- a/lib/pavilion/plugins/commands/wait.py
+++ b/lib/pavilion/plugins/commands/wait.py
@@ -75,10 +75,10 @@ class WaitCommand(commands.Command):
                 if time.time() > (start_time + 5*periodic_status_count):
                     for test in tmp_statuses:
                          stat = [str(time.ctime(time.time())), ':',
-                               str(test['test_id']),
-                               test['name'],
-                               test['state'],
-                               test['note']]
+                                 str(test['test_id']),
+                                 test['name'],
+                                 test['state'],
+                                 test['note']]
                          fprint(' '.join(stat))
                     periodic_status_count += 1
 
--- a/lib/pavilion/plugins/commands/wait.py
+++ b/lib/pavilion/plugins/commands/wait.py
@@ -74,7 +74,8 @@ class WaitCommand(commands.Command):
             if not args.silent:
                 if time.time() > (start_time + 5*periodic_status_count):
                     for test in tmp_statuses:
-                        stat = [str(time.ctime(time.time())), ':',
+                        stat = [str(time.ctime(time.time())), ':', 
+                                'test #',
                                 str(test['test_id']),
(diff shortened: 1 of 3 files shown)


lib/pavilion/plugins/commands/wait.py
```diff
@@ -49,8 +56,7 @@ def run(self, pav_cfg, args):
 
         final_statuses = 0
 
-        while (final_statuses < len(tmp_statuses)) and \
-              ((time.time() - start_time) < float(args.timeout)):
+        while final_statuses < len(tmp_statuses):
```
Paul-Ferrell: This would be greatly simplified by defaulting --timeout to None (no timeout), and dropping the no_timeout argument.

```python
end_time = None
if args.timeout is not None:
    end_time = time.time() + args.timeout
while (final_statuses < len(tmp_statuses) and
        (end_time is None or time.time() < end_time)):
```

lib/pavilion/plugins/commands/wait.py
```diff
@@ -30,17 +30,24 @@ def _setup_arguments(self, parser):
...
-        parser.add_argument(
-            '-t', '--timeout', action='store', default='60',
-            help='Maximum time to wait for results in seconds. Default=60.'
-        )
         parser.add_argument(
             'tests', nargs='*', action='store',
             help='The name(s) of the tests to check.  These may be any mix of '
                  'test IDs and series IDs.  If no value is provided, the most '
                  'recent series submitted by this user is checked.'
         )
```
Paul-Ferrell: I'm glad you're getting this deep into the argument parser library.

Paul-Ferrell:
We'll look at this at the meeting. It's fine to merge if this is the behavior we want.

Paul-Ferrell:
- Wait forever by default
 - Give periodic pav status (one liner) by default (add --silent)

lib/pavilion/plugins/commands/wait.py
```diff
@@ -40,17 +42,25 @@ def _setup_arguments(self, parser):
...
     def run(self, pav_cfg, args):
-        # Store the initial time for timeout functionality.
         start_time = time.time()
 
         tmp_statuses = status.get_statuses(pav_cfg, args, self.errfile)
 
         final_statuses = 0
 
-        while (final_statuses < len(tmp_statuses)) and \
-              ((time.time() - start_time) < float(args.timeout)):
```
Paul-Ferrell: I argued with Nick about doing it this way. The status is for informational purposes only. The list of statuses is subject to change, and is generally not reliable. It even says so in their documentation.

The right way to tell if a test is done isn't to check its status, it's to check use the test.wait() method. It's a slightly harder problem to do this for multiple tests, but not bad. I'd appreciate it if you'd fix it the right way. 

Sorry I didn't tell you this in the first place. I forgot how it was implemented.

lib/pavilion/plugins/commands/wait.py
```diff
@@ -60,6 +70,19 @@ def run(self, pav_cfg, args):
 
             tmp_statuses = status.get_statuses(pav_cfg, args, self.errfile)
 
+            # print status every 5 seconds
```
Paul-Ferrell: Why not use the status command bits to just print the status table as normal?

francinelapid: It will reprint the table each time, headers included:

```
 Test statuses                                                   
---------+---------------+---------+----------+--------------------
 Test id | Name          | State   | Time     | Note               
---------+---------------+---------+----------+--------------------
 11      | test.basic | RUNNING | 16:49:57 | Currently running. 

 Test statuses                                                   
---------+---------------+---------+----------+--------------------
 Test id | Name          | State   | Time     | Note               
---------+---------------+---------+----------+--------------------
 11      | test.basic | RUNNING | 16:49:57 | Currently running. 
```

lib/pavilion/plugins/commands/wait.py
francinelapid: If I use the `test.wait()` method, then I won't be able to periodically print the test statuses while waiting for the results. 

Instead of checking the statuses, I think I'll check for the `RUN_COMPLETE` file, which is how `test.wait` checks if a test is finished.

PR #134, opened 2020-02-03
Table result parser typo and nested dictionary customization

francinelapid:
TODO:

- [x] faulty conditional (`if len(col_names) != col_num`)
- [x] option to let user choose which one (row/col) will be nested in the dictionary
- [x] add test to verify row/col switch is still accurate
- [x] make `by_column` default `False`


lib/pavilion/plugins/results/table.py
```diff
@@ -42,7 +42,7 @@ def _check_args(self, delimiter=None, col_num=None, has_header=None,
                     col_names = []):
 
         if len(col_names) is not 0:
-            if len(col_names) is not col_num:
+            if len(col_names) is not int(col_num):
```
Paul-Ferrell: The is keyword isn't appropriate here. It's a pointer comparison:
```C
if ( &a == &b) ...
```

You see it used with `None` a lot because you're not checking if the value is `None` you're checking to see if the value is a pointer to the singleton `None` object (basically saying `&a == NULL`)

Paul-Ferrell: Though I should probably point out that the None object is obviously not literally located at NULL.

francinelapid: `if len(col_names) != int(col_num)`

lib/pavilion/plugins/results/table.py
```diff
@@ -28,21 +28,28 @@ def get_config_items(self):
...
             yc.StrElem(
                 'has_header', default='False',
-                help_text="Set True if there is a column for row names."
+                help_text="Set True if there is a column for row names. Will "
+                          "create dictionary of dictionaries."
             ),
             yc.ListElem(
                 'col_names', required=False, sub_elem=yc.StrElem(),
                 help_text="Column names if the user knows what they are."
+            ),
```
Paul-Ferrell: Would you be ok with making this the default (and thus probably renaming it 'by_column' or something instead). It breaks backwards compatability, but that's ok.

lib/pavilion/plugins/results/table.py
```diff
@@ -27,33 +27,41 @@ def get_config_items(self):
...
+            yc.StrElem(
+                'by_column', choices=['True', 'False'], default='True',
+                help_text="Set to True if the user wants to organize the "
+                          "nested dictionaries by columns. Default False. "
```
Paul-Ferrell: Instead of ignoring, validate this in the check_args method.

Issue #121, opened 2020-01-27
Cat command inserts extra newlines


Paul-Ferrell:
When you cat files extra newlines are inserted (one per block of the file read).

You should add a test that makes sure the contents of the file and the output of the command are the same.

j-ogas:
Closing, resolved via #135.

PR #135, opened 2020-02-04
fix cat

j-ogas:
WIP.

Addresses #121.

TODO:
- [x] update test target to generic file without user specific data
- [x] make travis happy

### DIFF
--- a/lib/pavilion/plugins/commands/cat.py
+++ b/lib/pavilion/plugins/commands/cat.py
@@ -19,7 +19,8 @@ class FileCommand(commands.Command):
             short_help="Print file information of <job id>"
         )
 
-    def _setup_arguments(self, parser):
+    @staticmethod
+    def _setup_arguments(parser):
         parser.add_argument(
             'job_id', type=int,
             help="job id",
@@ -44,28 +45,27 @@ class FileCommand(commands.Command):
                           file=sys.stderr, color=output.RED)
             return errno.EEXIST
 
-        return print_file(job_dir/args.file)
+        return self.print_file(job_dir/args.file)
 
+    def print_file(self, file):
+        """Print the file at the given path.
+        :param path file: The path to the file to print.
+        """
 
-def print_file(file):
-    """Print the file at the given path.
-    :param path file: The path to the file to print.
-    """
+        try:
+            with file.open() as file:
+                while True:
+                    block = file.read(4096)
+                    if not block:
+                        break
+                    output.fprint(block, file=self.outfile, end="")
 
-    try:
-        with file.open() as file:
-            while True:
-                block = file.read(4096)
-                if not block:
-                    break
-                print(block, file=sys.stdout, end="")
+        except IsADirectoryError:
+            output.fprint("{} is a directory.".format(file), sys.stderr,
+                          color=output.RED)
+            return errno.EINVAL
 
-    except IsADirectoryError:
-        output.fprint("{} is a directory.".format(file), sys.stderr,
-                      color=output.RED)
-        return errno.EINVAL
-
-    except FileNotFoundError:
-        output.fprint("file '{}' does not exist.".format(file), sys.stderr,
-                      color=output.RED)
-        return errno.EEXIST
+        except FileNotFoundError:
+            output.fprint("file '{}' does not exist.".format(file), sys.stderr,
+                          color=output.RED)
+            return errno.EEXIST
--- a/lib/pavilion/plugins/commands/cat.py
+++ b/lib/pavilion/plugins/commands/cat.py
@@ -54,11 +54,8 @@ def print_file(file):
 
     try:
         with file.open() as file:
-            while True:
-                block = file.read(4096)
-                if not block:
-                    break
-                print(block, file=sys.stdout, end='')
+            data = file.read()
+            output.fprint(data, file=sys.stdout)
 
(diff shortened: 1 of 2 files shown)


test/tests/cat_tests.py
```diff
@@ -26,10 +26,22 @@ def test_cat(self):
...
+#!/bin/bash
+
+# The first (and only) argument of the build script is the test id.
+export TEST_ID={}
+export PAV_CONFIG_FILE={}
+source {}
+
+# Perform the sequence of test commands.
+echo "Hello World."
+""".format(test.id, test.TODO, test.TODO)
```
j-ogas: @Paul-Ferrell Where can I get the format field values from the quick test object?

test/tests/cat_tests.py
```diff
@@ -26,10 +26,22 @@ def test_cat(self):
...
 
         for arg_set in arg_sets:
             args = arg_parser.parse_args(arg_set)
             cat_cmd.run(self.pav_cfg, args)
-            output.fprint(cat_cmd.outfile)
-            output.fprint(cat_cmd.errfile)
+            # FIXME: always true
+            if cat_cmd.outfile != true_out:
+                return errno.ENOMSG
```
j-ogas: Test always passes with this line.

Paul-Ferrell: cat_cmd.outfile is a file-like object. You have to read it. `cat_cmd.outfile.read()`

test/tests/cat_tests.py
```diff
@@ -26,10 +26,22 @@ def test_cat(self):
         cat_cmd.errfile = io.StringIO()
 
         arg_parser = arguments.get_parser()
-        arg_sets = (['cat', str(test.id), 'build.sh'],)
```
Paul-Ferrell: To get a nice long file to mess with: 

```python
test_cfg = self._quick_test_cfg()
test_cfg['run']['cmds'] = ['echo "hello world"'] * 1000
test = self._quick_test(test_cfg)
```

test/tests/cat_tests.py
Paul-Ferrell: Use the test object assert statements. 
```python
self.assertEqual(thing1, thing2)
#In this case:
true_out = (test.path/arg_set[-1]).open().read()
self.assertEqual(cat_cmd.outfile.read(), true_out)
```

lib/pavilion/plugins/commands/cat.py
```diff
@@ -54,11 +54,8 @@ def print_file(file):
 
     try:
         with file.open() as file:
-            while True:
```
Paul-Ferrell: The way you did this before was better; you don't want to read a whole arbitrary file into memory at once.
Just add `end=""` to your print statement. (fprint passes extra keyword arguments to print())

lib/pavilion/plugins/commands/cat.py
```diff
@@ -58,7 +58,7 @@ def print_file(file):
                 block = file.read(4096)
                 if not block:
                     break
-                output.fprint(block, file=sys.stdout)
+                print(block, file=sys.stdout, end="")
```
Paul-Ferrell:
```
output.fprint(block, file=self.outfile, end="")
```

test/tests/cat_tests.py
```diff
@@ -19,17 +19,21 @@ def teatDown(self):
...
         cat_cmd.outfile = io.StringIO()
         cat_cmd.errfile = io.StringIO()
 
         arg_parser = arguments.get_parser()
-        arg_sets = (['cat', str(test.id), 'build.sh'],)
-
+        arg_sets = (['cat', str(test.id), 'run.tmpl'],)
         for arg_set in arg_sets:
             args = arg_parser.parse_args(arg_set)
+            sys.stdout = cat_cmd.outfile
```
Paul-Ferrell: no!!!

lib/pavilion/plugins/commands/cat.py
```diff
@@ -44,28 +45,27 @@ def run(self, pav_cfg, args):
...
+                    output.fprint(block, file=self.outfile, end="")
 
-    try:
-        with file.open() as file:
-            while True:
-                block = file.read(4096)
-                if not block:
-                    break
-                output.fprint(block, file=sys.stdout)
+        except IsADirectoryError:
```
Paul-Ferrell: You should also catch OSError, IOError, and PermissionError. You can get them as a group, and then give a generic message like:
"Error reading file '{}': {}".format(file_path, err)

lib/pavilion/plugins/commands/cat.py
```diff
@@ -44,28 +45,25 @@ def run(self, pav_cfg, args):
...
-
-    except FileNotFoundError:
-        output.fprint("file '{}' does not exist.".format(file), sys.stderr,
-                      color=output.RED)
-        return errno.EEXIST
+        except (FileNotFoundError,
+                IsADirectoryError,
+                OSError,
+                PermissionError) as err:
+            output.fprint("Error reading file '{}': {}".format(file), err,
```
Paul-Ferrell: :man_facepalming: - I should be more explicit. For generic errors that you don't know the cause for (OSError, IOError), give a generic message. For errors with an obvious cause, give a _specific_ mesage like you were doing before. 

More specific errors are almost always better. :)

Issue #110, opened 2020-01-08
Use threads to parallelize builds on the frontend.


Paul-Ferrell:
When running a bunch of tests, Pavilion builds each of them sequentially. We should instead 
run them in parallel.

This will require learning how to write a threaded Python program, which will be fun.

PR #137, opened 2020-02-05
Concurrent builds

francinelapid:
#110 

Not quite parallel, but concurrent. 

TODO: 
- [x] limit number of threads executing at the same time
- [x] mechanism to kill threads
- [ ] writing to errfile should be atomic (use locks)

### DIFF
--- a/lib/pavilion/plugins/commands/run.py
+++ b/lib/pavilion/plugins/commands/run.py
@@ -2,6 +2,7 @@ import codecs
 import errno
 import pathlib
 import time
+import threading
 from collections import defaultdict
 
 from pavilion import commands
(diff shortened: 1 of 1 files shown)


lib/pavilion/plugins/commands/run.py
```diff
@@ -18,6 +19,18 @@
 from pavilion.test_run import TestRun, TestRunError
 
 
+def build_test_thread(test, failed_list, err_out):
+
+    if not test.build():
+        fprint("Error building test: ", file=err_out, color=output.RED)
```
Paul-Ferrell: You should surround all the output with a lock, to make sure it doesn't get interleaved.

lib/pavilion/plugins/commands/run.py
```diff
@@ -18,6 +19,18 @@
...
+        fprint("status {status.state} - {status.note}"
+               .format(status=test.status.current()),
+               file=err_out)
+        fprint("For more information, run 'pav log build {}'"
+               .format(test.id, file=err_out))
+        failed_list.append(test)
```
Paul-Ferrell: I don't know if you checked, but this operation is thread safe. :+1:

francinelapid:
Closed because Paul's branch that separates all build-related functionality includes this

PR #155, opened 2020-02-26
catch empty suite, empty test_cfg, and test list missing '-f'

CalvinDSeamons:
Three Issues were resulting in tracebacks: 
1) Running a `test.yaml` that was completely empty. Caught with raised error.
2) Running a `test.yaml` that is a list of tests (such as postDST.yaml) without the `-f` flag.
    Caught with a Attribute Error Exception. 
3) While not technically an error having `test.yaml` with only `base:` would run and just not do anything. If this happens more than once tracebacks will happen in `resolve_permutations`. Also, I feel the user should be told they are running a test that literally does nothing.

### DIFF
--- a/lib/pavilion/test_config/setup.py
+++ b/lib/pavilion/test_config/setup.py
@@ -349,21 +349,32 @@ def resolve_inheritance(base_config, suite_cfg, suite_path):
     # A list of tests whose parent's have had their dependencies
     # resolved.
     ready_to_resolve = list()
-    for test_cfg_name, test_cfg in suite_cfg.items():
-        if test_cfg.get('inherits_from') is None:
-            test_cfg['inherits_from'] = '__base__'
-            # Tests that depend on nothing are ready to resolve.
-            ready_to_resolve.append(test_cfg_name)
-        else:
-            depended_on_by[test_cfg['inherits_from']].append(test_cfg_name)
+    if suite_cfg is None:
+        raise TestConfigError("Test Suite {} is empty.".format(suite_path))
+    try:
+        for test_cfg_name, test_cfg in suite_cfg.items():
+            if test_cfg is None:
+                raise TestConfigError("{} in {} is empty. Nothing will execute."
+                                      .format(test_cfg_name, suite_path))
+            if test_cfg.get('inherits_from') is None:
+                test_cfg['inherits_from'] = '__base__'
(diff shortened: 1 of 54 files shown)


francinelapid:
For # 2 in your description above, how are you running that list of tests?

`pav run test`
or
`pav run -f test.yaml`

(if test.yaml is just a list of tests)

CalvinDSeamons:
I run it using a `pav run test-list`.  This results in a traceback error which i don't think Paul wants.  The command `pav run -f test.yaml`  is fine as that is the proper syntax. Either way it catches a chain of errors. We can talk today about a better message to return. Sorry for the confusion.

lib/pavilion/test_config/setup.py
```diff
@@ -349,21 +349,32 @@ def resolve_inheritance(base_config, suite_cfg, suite_path):
...
+                depended_on_by[test_cfg['inherits_from']].append(test_cfg_name)
 
+            try:
+                suite_tests[test_cfg_name] = TestConfigLoader().normalize(
+                    test_cfg)
+            except (TypeError, KeyError, ValueError) as err:
+                raise TestConfigError(
+                    "Test {} in suite {} has an error: {}"
+                    .format(test_cfg_name, suite_path, err))
+    except AttributeError as err:
```
Paul-Ferrell: That postDST.yaml file shouldn't be there, and isn't in the correct format for '-f' anyway.

lib/pavilion/test_config/setup.py
```diff
@@ -349,21 +349,32 @@ def resolve_inheritance(base_config, suite_cfg, suite_path):
...
-            suite_tests[test_cfg_name] = TestConfigLoader().normalize(test_cfg)
-        except (TypeError, KeyError, ValueError) as err:
-            raise TestConfigError(
-                "Test {} in suite {} has an error: {}"
-                .format(test_cfg_name, suite_path, err))
+    if suite_cfg is None: # Catch null test suites.
+        raise TestConfigError("Test Suite {} is empty.".format(suite_path))
+    try:
+        for test_cfg_name, test_cfg in suite_cfg.items():
+            if test_cfg is None:
```
Paul-Ferrell: I think this is actually a bug in yaml_config, but this is all a reasonable fix for now.

Paul-Ferrell:
I think this is actually a reasonable fix.

Issue #140, opened 2020-02-06
Add a 'files' directive, to directly define files in the config.


Paul-Ferrell:
This is for dynamically creating small files on the fly. 

```yaml
mytest:
  variables:
    xdim: 104
    ydim: 104
    zdim: 104
    runtime: 10
  build:
    files:
      'datfiles/hpgc.dat':
         - 'HPCG benchmark input file'
         - 'Sandia National Laboratories; University of Tennessee, Knoxville'
         - '{{xdim}} {{ydim}} {{zdim}}'
         - '{{runtime}}'
```

 - The file path relative to the build directory.
 - The path must be within the build directory.
 - Directories in the path should be created by Pavilion, if necessary.
 - The generated file must be included in the build hash.
 - The 'files' directive is allowed in both the 'build' and 'run' sections.
 - If this is in the 'build' section, no deferred variables are allowed.

PR #166, opened 2020-03-09
add build-time file creation

j-ogas:
Partially Addresses #140 (build phase portion)

TODO:
- [x] add a new yc class
- [x] add build phase file creation
- [x] add hash
- [x] add documentation
- [x] add unit tests
- [x] prevent creating files outside the top-level build directory
- [x] make tests pass


lib/pavilion/test_config/file_format.py
```diff
@@ -19,6 +19,19 @@ class TestConfigError(ValueError):
...
+    _NAME_RE = re.compile(r"^[\w\-. ]+$")
+
+    # TODO: Not sure what to alter here.
+    def __init__(self, name=None, **kwargs):
+        super(FlexibleKeyElem, self).__init__(name=name,
+                                              sub_elem=yc.ListElem(),
+                                              defaults=None,
+                                              **kwargs)
+
+
```
j-ogas: Not sure what to do with this; specifically I'm not sure how I can override `CategoryElem` to give me a list of dictionaries where the `key` is the file name and the `value` is a list of strings (body of a file).

Any advice @Paul-Ferrell ?

Paul-Ferrell: You don't have to modify __init__ at all.

I'd name it 'PathCategory', set the _NAME_RE to something that will match paths, and that's it. 
You'll also want to preserve the case of the keys. To do that, when you instantiate it, set key_case=PathCategory.KC_MIXED

lib/pavilion/builder.py
```diff
@@ -596,6 +596,10 @@ def _setup_build_dir(self, dest):
                         "Could not copy test src '{}' to '{}': {}"
                         .format(src_path, dest, err))
 
+        # Generate file(s) from build_config
+        if self._config.get('make_files'):
+            self.make_files(self._config.get('make_files', []), dest)
```
Paul-Ferrell: 'make_files' is a dict, not a list, and this get should default to a dict.

lib/pavilion/builder.py
```diff
@@ -608,6 +612,16 @@ def _setup_build_dir(self, dest):
...
 
+    def make_files(self, files, dst):
+        for file in files:
+            # Assumes that each file is a path, e.g., 'somedir/file', or './file'
+            dirname = os.path.dirname(file)
+            Path(dst / dirname).mkdir(parents=True, exist_ok=True)
+            path = Path(file)
+            with open(dst / path, 'w') as f:
+                for line in file:
+                    f.write(line)
```
Paul-Ferrell: I'm pretty sure you're writing the filename to file here, not the lines in the file.

lib/pavilion/builder.py
```diff
@@ -608,6 +612,16 @@ def _setup_build_dir(self, dest):
                     "Could not copy extra file '{}' to dest '{}': {}"
                     .format(path, dest, err))
 
+    def make_files(self, files, dst):
```
Paul-Ferrell: dst is the build directory, which is also self.path.

lib/pavilion/builder.py
```diff
@@ -608,6 +612,16 @@ def _setup_build_dir(self, dest):
...
+        for file in files:
```
Paul-Ferrell: Files will be a dict of lists of strings: `{'foo/bar': ['line 1', 'line 2']}`. 
You can iterate over both the keys and values with:
```
for filename, content in files.items():
  ...
```

lib/pavilion/builder.py
```diff
@@ -608,6 +612,16 @@ def _setup_build_dir(self, dest):
...
+            # Assumes that each file is a path, e.g., 'somedir/file', or './file'
+            dirname = os.path.dirname(file)
```
Paul-Ferrell: What happens if the file already exists? 
It should probably only be an error if you're trying to overwrite a directory with a regular file. 
All the cases still have to be handled though.

lib/pavilion/builder.py
Paul-Ferrell: The logic of writing a file to a file object should be its own method.
 1. It will make it easier to write tests of that 'unit'. 
 2. You're going to have to generate these files in two places. Once to hash them under create_build_hash, and once to actually write them once we're creating the build.

The logic of looping over the files can be live outside of it.

docs/tests/build.rst
```diff
@@ -103,13 +105,38 @@ When downloading source, we by default use the last of the url path as
...
+overwritten.
+
+.. code-block:: yaml
+
+    mytest:
+      build:
+        source_location: mytest.zip
+        cmds: 'make'
+        make_files:
+          # Files created this way will be actual files, not symlinks.
```
Paul-Ferrell, line 122: That's not true. These still go into the build directory during the build. When the build is 'copied' to the test_run, they'll end up being symlinks.

lib/pavilion/builder.py
```diff
@@ -209,6 +212,15 @@ def __init__(self, pav_cfg, test, mb_tracker, build_name=None):
...
         self.fail_path = pav_cfg.working_dir/'builds'/fail_name
 
+        # Don't allow syntax that may cause a file to be written outside of the
+        # build context directory.
+        files_to_create = self._config.get('create_files')
+        if files_to_create:
+            for file, contents in files_to_create.items():
+                if '../' in str(file):
+                    raise TestBuilderError("'create_file: {}': dangerous syntax"
+                                           "'..'".format(file))
```
Paul-Ferrell, line 222: Checking for '../' isn't good enough. You also need to check that the path doesn't start with a '/', because 
combining '/foo' with '/bar' gets you '/bar', not '/foo/bar'.

Toss this into utils.py:
```python
def dir_contains(file, directory):
    """Check if 'file' is or is contained by 'directory'."""
    file = pathlib.Path(file).resolve()
    directory = pathlib.Path(directory).resolve()
    while file.parent != file:
        if file == directory:
            return True
        file = file.parent
    return False
```

lib/pavilion/builder.py
```diff
@@ -258,13 +272,27 @@ def create_build_hash(self):
...
 
+        # Hash created build files. These files are generated at build time in
+        # the test's build directory but we need the contents of these files
+        # hashed before build time. Thus, we include a hash of each file
+        # consisting of the filename (including path) and it's contents via
+        # IOString object.
+        files_to_create = self._config.get('create_files')
+        if files_to_create:
+            for file, contents in files_to_create.items():
+                io_contents = io.StringIO()
```
Paul-Ferrell, line 289: You can use `with` with io.StringIO().  You don't have to though.

test/tests/builder_tests.py
```diff
@@ -109,6 +110,36 @@ def test_setup_build_dir(self):
             self._cmp_files(test_archives/file,
                             test.builder.path/file)
 
+    def test_create_file(self):
```
Paul-Ferrell, line 113: You need to test that you can't create a file outside of the build directory, and make sure you don't overrwrite a directory with a file.

lib/pavilion/builder.py
```diff
@@ -607,6 +643,22 @@ def _setup_build_dir(self, dest):
                         "Could not copy test src '{}' to '{}': {}"
                         .format(src_path, dest, err))
 
+        # Create build time file(s).
+        files_to_create = self._config.get('create_files')
+        if files_to_create:
+            for file, contents in files_to_create.items():
+                file_path = Path(utils.resolve_path(dest / file))
+                # Do not allow file to clash with existing directory.
+                if os.path.isdir(str(file_path)):
```
Paul-Ferrell, line 652: file_path.is_dir()

lib/pavilion/builder.py
```diff
@@ -607,6 +643,22 @@ def _setup_build_dir(self, dest):
...
+                    raise TestBuilderError("'create_file: {}' clashes with"
+                                           " existing directory in test source."
+                                           .format(str(file_path)))
+                dirname = file_path.parent
+                Path(dest / dirname).mkdir(parents=True, exist_ok=True)
```
Paul-Ferrell, line 657: `(dest/dirname).mkdir(parents=True, exist_ok=True)`

(dest/dirname) is a Path object.

lib/pavilion/builder.py
```diff
@@ -607,6 +643,22 @@ def _setup_build_dir(self, dest):
...
+                with open(str(file_path), 'w') as file_:
```
Paul-Ferrell, line 658: `with file_path.open('w') as file_:`

lib/pavilion/test_config/file_format.py
```diff
@@ -19,6 +19,11 @@ class TestConfigError(ValueError):
 VAR_NAME_RE = re.compile(r'^[a-zA-Z][a-zA-Z0-9_-]*[?+]?$')
 
 
+class PathElem(yc.CategoryElem):
+    """This is for category elements that need a valid unix path regex."""
+    _NAME_RE = re.compile(r"([^ !$`&*()+]|(\\[ !$`&*()+]))+")
```
Paul-Ferrell, line 24: As per my comments in chat, this can just be r".+$". There are some corner cases, but nothing worth checking for..

PR #183, opened 2020-03-20
Doc fix

Paul-Ferrell:
This is almost all documentation fixes, mostly to links. There are very few changes to actual content. 

The one big addition is in the unit tests. We now check all the links, both internal and external, in all of the generated documentation.


docs/basics.rst
```diff
@@ -171,7 +171,7 @@ From the above, you may have noticed that each test gets a series id
 like ``s24`` and a test id like ``41``. You can use these id's to
 reference tests or suites of tests to get their status, results, and
 logs through the pavilion interface. The ID's are unique for a given
-Pavilion `working\_directory <config.html#working_dir>`__, but they will
+Pavilion `working\_directory <config.html#working-dir>`__, but they will
```
francinelapid: Shouldn't this be `config.rst`? I don't see a `config.html` in the `docs` directory.

francinelapid: Shouldn't this be `config.rst`? I don't see a `config.html` in the docs section.

docs/config.rst
```diff
@@ -4,9 +4,9 @@ Configuring Pavilion
 Pavilion is driven largely by configurations. This documentation page covers
 the ``pavilion.yaml`` file, which sets global pavilion settings.
 
-See `Test Configs <tests/basics.html>`__,
-`Host Configs <tests/basics.html#host-configs>`__,
-`Mode Configs <tests/basics.html#mode-configs>`__, and
+See `Test Configs <tests/format.html>`__,
+`Host Configs <tests/format.html#host-configs>`__,
+`Mode Configs <tests/format.html#mode-configs>`__, and
```
francinelapid: Should all of these be `.rst`s?

docs/basics.rst
Paul-Ferrell: They're automatically generated when you run make html in the docs directory, but they aren't in the repo. The relative locations are the same, so the paths still work.

PR #188, opened 2020-03-24
pav clean improvements

### LINKED ISSUE #172: Improve Pav clean arg parsing
Right now it fails for things like '10 days' (it does 1 day) because of bad parsing.

Add the ability to handle older-than minutes and hours.

Add a --all option.

francinelapid:
- [x] add `--all` flag
- [x] improve date parsing
- [x] `--older-than='3 weeks'`
- [x] add `minutes`
- [x] add `hours`

Closes #172


lib/pavilion/plugins/commands/clean.py
```diff
@@ -215,10 +221,12 @@ def get_month_delta(months):
     cur_year = today.year
     cur_day = today.day
     cur_month = today.month
-    cur_time = today.time
+    cur_time = today.time().hour
 
     if cur_month - months <= 0:
```
Paul-Ferrell: All of this should be done with the datetime timedelta object.

Need now - 12 months?
```python
import datetime as dt
months = 12
dt.datetime.now() - dt.timedelta(days=30*months)
```

Paul-Ferrell: 30 days is a reasonable assumption for the length of a month. It's almost right 100% the time, depending on your definition of almost.

lib/pavilion/plugins/commands/clean.py
```diff
@@ -102,7 +108,7 @@ def run(self, pav_cfg, args):
                     build_origin_symlink.resolve().exists()):
                 build_origin = build_origin_symlink.resolve()
 
-            if test_time < cutoff_date:
```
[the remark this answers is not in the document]
francinelapid: But pizza _is_ pie...

lib/pavilion/plugins/commands/clean.py
francinelapid: deleted this whole function because with the new implementation, it only needs ~2 lines

lib/pavilion/plugins/commands/clean.py
```diff
@@ -46,7 +53,8 @@ def run(self, pav_cfg, args):
                 cutoff_date = datetime.today() - timedelta(
                     weeks=int(args.older_than[0]))
             elif 'month' in args.older_than or 'months' in args.older_than:
-                cutoff_date = get_month_delta(int(args.older_than[0]))
```
Paul-Ferrell: I forgot about this, but Kody's code is super wrong.

Here's a snippet I wrote to make sure:
```python
#!/usr/bin/python3

import argparse
import sys

p = argparse.ArgumentParser()
p.add_argument('--older-than', nargs='+', action='store')

args = p.parse_args()

print(args.older_than)
```

Paul-Ferrell: As you'll see when you play with this, args.older_than is a list (because of nargs='+'), which allows for getting multiple values for this argument (which we don't want). Since the action is 'store' though, additional values given just overwrite the original list. 

This whole section doesn't work at all because when you look for 'days' in args.older_than, it will find stuff like '5 days' and will never match. If it did match, it would try to convert '5 days' into an int, which is a ValueError.

PR #194, opened 2020-03-31
mpirun-slurm plugin improvements

francinelapid:
TO DO:
- [x] `--map-by ppr:<procs_per_node>:node`
- [x] `--rank-by`
- [x] `--bind-to`
- [x] `-mca`
- [ ] `-np`
- [x] num_nodes < alloc nodes


lib/pavilion/plugins/sched/slurm.py
```diff
@@ -307,6 +307,73 @@ def slurm_states(state):
     return states
 
 
+def get_slurm_elems():
```
Paul-Ferrell, line 310: You've got the right idea here, but make keep this as a private, static '_get_config_elems()' method instead.

PR #206, opened 2020-04-08
wait command fixes

francinelapid:
TODO:

- [x] add `--oneline` flag to wait output
- [x] make sure config is written to atomically

`pav wait --oneline` will produce:
`Wed Apr  8 14:40:04 2020: 10(test name) RUNNING | 11(test name) RUNNING | 12(test name) RUN_TIMEOUT | `


lib/pavilion/test_run.py
```diff
@@ -412,15 +412,26 @@ def _save_config(self):
 
         config_path = self.path/'config'
 
+        # make lock
+        lock_path = self.path/'config.lockfile'
+        config_lock = lockfile.LockFile(
+            lock_path,
+            group=self._pav_cfg.shared_group
+        )
+
```
Paul-Ferrell, line 421: There's a much easier way:
https://rcrowley.org/2010/01/06/things-unix-can-do-atomically.html

Paul-Ferrell: Nope, I'm wrong. If you were just creating the file, not replacing it, writing it to a tmp file and `os.rename`ing it is atomic. But since you have to delete the original first, it isn't. So you do have to use a lock.

lib/pavilion/test_run.py
```diff
@@ -412,15 +412,26 @@ def _save_config(self):
...
         try:
+            config_lock.lock()
```
Paul-Ferrell, line 423:
```python
with config_lock:
   # do stuff
```
or 

```python
try: 
  config_lock.lock()
  # do stuff
finally:
  config_lock.unlock()
```

lib/pavilion/test_run.py
```diff
@@ -412,15 +412,26 @@ def _save_config(self):
...
+                .format(self.name, self.path, err))
         except TypeError as err:
-            raise TestRunError("Invalid type in config for ({}): {}"
-                               .format(self.name, err))
+            raise TestRunError(
+                "Invalid type in config for ({}): {}"
+                .format(self.name, err))
 
     @classmethod
     def _load_config(cls, test_path):
```
Paul-Ferrell, line 437: The problem also includes reading the config file at an unsafe time. You need to acquire the same lock when loading the config too.

PR #208, opened 2020-04-09
Added sched vars to results

### LINKED ISSUE #205: Add scheduler variables to all results.
We need to be logging exactly what nodes each test ran on, and I think the easiest way to do that is to log all of the scheduler variables in each test run's results.

This will be trivial to implement.

Paul-Ferrell:
Closes #205


Paul-Ferrell:
I'll add that, along with something else I've been wanting to add...

________________________________

PR #210, opened 2020-04-09
Autogenerate subtitle based on permutations.

### LINKED ISSUE #209: A subtitle should be auto-generated for permuted tests.
Permuted tests should automatically get a subtitle, if one isn't defined, based on permuted variables.

Paul-Ferrell:
Fixes #209


lib/pavilion/test_run.py
```diff
@@ -407,6 +402,39 @@ def run_cmd(self):
...
+                        shortest = ''
+                        for subval in val.values():
+                            if not shortest or len(subval) < len(shortest):
+                                shortest = subval
+                        sub_parts.append(shortest)
+                    else:
+                        sub_parts.append(val)
+            subtitle = '-'.join(sub_parts)
+
+        if subtitle is not None:
```
nicholas-sly, line 433: The logic here seems to cover:
- There is no provided subtitle but at least one variable is being permuted over.
- There is a provided subtitle, but no variables being permuted over.

It does not seem to cover if there is a provided subtitle and there is at least one variable being permuted over.

Did I miss that somewhere?

Paul-Ferrell: Hold off on this one. I don't like how I calculate the key, and I want team input.

RELEASE.txt
```diff
@@ -4,6 +4,12 @@ RELEASE=2.1.2
 
 # Release History
 
+## 2.2 Pre-release notes
+ - Backwards compatibility breaks
+   - A 'subtitle' is now required for all permuted tests.
+   - A duplicate keys in configs are now detected and cause errors.
```
nicholas-sly, line 10: 'All|Any duplicate keys' or 'A duplicate key'.

lib/pavilion/test_config/resolver.py
```diff
@@ -671,12 +671,37 @@ def resolve_permutations(test_cfg, base_var_man):
...
+                    valid_subtitle = False
+                else:
+                    subtitle.append('{{' + per_var + '}}')
+
+            subtitle = '-'.join(subtitle)
+
+            if valid_subtitle:
+                test_cfg['subtitle'] = subtitle
+            else:
+                raise TestConfigError(
```
nicholas-sly, line 700: I guess you decided not to just raise a warning.  I'm still not sure an exception is the right way to go.  A lack of a subtitle is inconvenient, but I don't think it means the tests shouldn't run.  It just means you'll have to go into the results or the run directory to figure out which one each one is.

Paul-Ferrell: We'll talk about this at the meeting today.

RELEASE.txt
```diff
@@ -4,6 +4,12 @@ RELEASE=2.1.2
 
 # Release History
 
+## 2.2 Pre-release notes
+ - Backwards compatibility breaks
+   - A 'subtitle' is now required for all permuted tests.
```
nicholas-sly, line 9: I guess this isn't true any longer, right?

Paul-Ferrell: damnit jim.

Issue #193, opened 2020-03-31
Add `.available()` method to scheduler plugins


Paul-Ferrell:
Scheduler plugins need to be able to check if the given scheduler is even available to use. 

Requirements:
 - Should be added to the base class and just raise NotImplementedError
 - Should be added as a method to all current schedulers. 
 - Should be checked before we start any test runs that use that scheduler.
  - The error and a reasonable message should be reported to the user.

PR #212, opened 2020-04-13
Check for unavailable schedulers

Paul-Ferrell:
Scheduler plugins must now implement an 'available()' method.
 - Test runs will fail when trying to run a test under an
   unavailable schedulers.
 - Added a unit test for this.


Paul-Ferrell:
Fixes #193

PR #223, opened 2020-04-28
command to check version of pavilion

CalvinDSeamons:
Pavilion command to check version of Pavilion. #148 

Right now the only way to executed the command is: `pav version` not `pav --version`.
I looked into making this a flag of `pav` but pay.py looked a little funky and I was conflicted on whether or not it made more sense to just have it as a command. 

I though an easy fix would be to have `aliases=['--version']` this results in an error even if the aliases is just a lowercase word like 'hello'. I'm not sure if the aliases feature is working as desired or if i just mucked it up.


lib/pavilion/plugins/commands/version.py
```diff
@@ -0,0 +1,40 @@
...
+        super().__init__(
+            name='version',
+            description='Display version of Pavilion.',
+            short_help="Displays the current version of Pavilion.",
+        )
+
+    def run(self, pav_cfg, args):
+        """Run fetches the version number from RELEASE.txt"""
+
+        version_path = str(pav_cfg.pav_root) + '/RELEASE.txt'
```
francinelapid, line 23: import pathlib
...
`version_path = pathlib.Path(pav_cfg.pav_root) / 'RELEASE.txt'`

Paul-Ferrell: pav_cfg.pav_root is already a pathlib object, so:
```
version_path = pav_cfg.pav_root / 'RELEASE.txt'
```

PR #227, opened 2020-05-05
new config elements for table result parser

francinelapid:
Added new config elements to help Pavilion narrow down where table might be located in output.

Phase 1: "Remove" unnecessary lines from file (in other words, only leave the relevant chunk)
- [x] `start_re`: Regex to help locate table
- [x] `nth_start_re`: nth occurrence of `start_re` to use
- [x] `start_skip`: number of lines between `start_re` and actual table* 
- [x] `row_num`: Number of rows including row of column names, if known

Phase 2: "Redraw" table
- [x] `ignore_rows`: list of rows indices Pavilion should ignore ignore
- [ ] `ignore_cols`: list of column indices Pavilion should ignore

Phase 3: Gather results into table
- leave existing code there

Other stuff to do:
- [x] update documentation
- [x] update regression tests

* actual table includes header (row with column names), if it exists


lib/pavilion/plugins/results/table.py
```diff
@@ -41,13 +41,27 @@ def get_config_items(self):
...
                           "Only set if `has_header` is True. "
                           "Otherwise, Pavilion will ignore."
+            ),
+            yc.StrElem(
+                'title', 
+                help_text="Partial or full title of table."
+            ),
+            yc.StrElem(
+                'row_num',
+                help_text="Number of row numbers, including column names."
```
Paul-Ferrell, line 51: What happens with tables with a dynamic number of rows? Do the lines skipped after the header count?

Also, "Number of rows, include the column header row."

lib/pavilion/plugins/results/table.py
```diff
@@ -41,13 +41,27 @@ def get_config_items(self):
                           "nested dictionaries by columns. Default False. "
                           "Only set if `has_header` is True. "
                           "Otherwise, Pavilion will ignore."
+            ),
+            yc.StrElem(
+                'title', 
+                help_text="Partial or full title of table."
```
Paul-Ferrell, line 47: What is this used for? 

(I'm sure that will be apparent through the code, but it needs to be apparent in the help_text too).

lib/pavilion/plugins/results/table.py
```diff
@@ -41,13 +41,27 @@ def get_config_items(self):
...
+            ),
+            yc.StrElem(
+                'row_num',
+                help_text="Number of row numbers, including column names."
+            ),
+            yc.StrElem(
+                'num_skip',
```
Paul-Ferrell, line 54: Skip the number of what?
'post_header_row_skip' is long, but obvious.

lib/pavilion/plugins/results/table.py
```diff
@@ -57,13 +71,37 @@ def _check_args(self, delimiter=None, col_num=None, has_header=None,
...
-                 has_header='', col_names=[], by_column=True):
+                 has_header='', col_names=[], by_column=True, 
+                 title=None, row_num=None, num_skip=None):
 
         match_list = []
+        lines = file.readlines()
+        new_lines = []
+        for line_index in range(len(lines)):
+            if title in lines[line_index]:
+                new_lines = lines[line_index:]
```
Paul-Ferrell, line 93: What if the title is more dynamic, or there just isn't one?

Paul-Ferrell:
Already addressed and merged.

PR #231, opened 2020-05-13
if skip, set run complete

CalvinDSeamons:
With Francine using some conditional test functionality she noticed that a `RUN_COMPLETE` file wasn't being placed in tests that were completed based on skips.


lib/pavilion/test_run.py
```diff
@@ -939,6 +939,7 @@ def _get_skipped(self):
             return False
         else:
             self.status.set(STATES.COMPLETE, matches)
```
Paul-Ferrell, line 941: I would recommend keeping the state at SKIPPED, and then setting run_complete.

PR #238, opened 2020-05-26
Status summary

CalvinDSeamons:
Status summary command. Usage looks like `pav status -s` or `pav status --summary`
Open to stylistic changes as this is a first draft. If you feeling strongly on adding more or less categories let me know! Output looks like: 

Screen Shot 2020-05-26 at 1 23 23 PM

Somethings to note right off the bat. The columns `passed` `failed` only apply to the `COMPLETED` row. Adding `n/a` or `none` to the other rows makes it look cluttered. Also I'm undecided on whether it should be titled `passed (of completed)` and `failed (of completed)` also having the total values not the percents such as `passed | 50% (10/20) |` etc. 

Still working on the possibility of a "animated progress bar" If that's something people want.

### DIFF
--- a/lib/pavilion/plugins/commands/status.py
+++ b/lib/pavilion/plugins/commands/status.py
@@ -224,8 +224,7 @@ def display_history(pav_cfg, args, outfile):
     return ret_val
 
 
-def print_summary(statuses, outfile, json=False):
-
+def print_summary(statuses, outfile):
     """Print_summary takes in a list of test statuses.
         It summarizes basic state output and displays
         the data to the user through draw_table.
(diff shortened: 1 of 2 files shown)


CalvinDSeamons:
I had some gnarly merge conflicts the `dbg_prints` got re-added i'll remove those in the next commit.

Paul-Ferrell:
I think it looks great. I wouldn't group the statuses though, it's a bit fragile if we add more. 
I'd just go by the statuses that are actually present. 

You can rely on 'FAILED', 'ERROR' and 'TIMEOUT' in the status names as indicators of 'bad' color, and make everything except complete a 'neutral' color.

CalvinDSeamons:
Moved to summary branch. The merge conflicts got so horrendous it was faster to just add the methods into a fresh master clone.

PR #240, opened 2020-05-26
Result Evaluations 104,150

### LINKED ISSUE #104: Result Analysis
Add a new layer that performs analysis on the results. It should allow for arbitrary (plugin) functions, and have some way to fold different parts of results objects into lists.

### LINKED ISSUE #150: Re-run result parsers on prior test run (with updated config).
It would be really useful if we could update the result parsers for a test and then re-run those on an existing test run. 

 - Check that only the result parsing section of the test config has changed.
 - Use the saved variable manager data to resolve the updated config.
 - The function that does this be part of the TestRun class.
 - This should be a sub-command of 'results'.
 - By default, don't log or save the new results.
 - With an option, save the updated config (and make sure the original (not intermediates in case of multiple reruns) is backed up). Also save the results (and back up the originals) and log the new results to the central results log (maybe a separate option).

Paul-Ferrell:
Fixes #104 
Fixes #150


Paul-Ferrell:
@francinelapid 
 1. The second test should pass, because `{{letters}}` is the same as `{{letters.0}}`, which resolves to `'b0 pass'`
 2. Both should fail with an error, because the 'result' key should no longer be settable from a result parser (it's now reserved for evaluations). I may have forgotten to add that bit of code.

francinelapid:
the `{{letters}}` is being permuted on, so one of the "b" tests would have 'b1 fail'
```
$ pavilion2 $ pav log run 36
b1 fail
```

Paul-Ferrell:
Oh, :P

Issue #84, opened 2019-12-12
pav show schedulers: add list of valid variables


j-ogas:
`pav show schedulers` variables should show a list of valid values that can be used.

PR #243, opened 2020-05-28
Added examples for sched vars.

Paul-Ferrell:
I couldn't come up with a good way to do this automatically, so now scheduler plugin writers have
to add examples manually.


Paul-Ferrell:
Fixes #84

PR #249, opened 2020-05-29
Summary Flag for pav status.

CalvinDSeamons:
This branch was formally submitted in the `status-summary` PR. I had some bad merge conflicts and it was easier to switch. I changed the table to be dynamic so now only states that exist in the test_statuses will show up. Table looks like:
Screen Shot 2020-05-29 at 11 24 44 AM

Only real change was adding the fractions to the `pass/fail` columns. I feel it's useful especially if you are running this on a large group of tests. That being said it doesn't look the best and a `" "`(space) gets wrapped in draw_tables so its a bit squished this way.


lib/pavilion/plugins/commands/status.py
```diff
@@ -321,3 +328,90 @@ def display_history(self, pav_cfg, args):
...
+
+        if 'COMPLETE' in summary_dict.keys():
+            fails = summary_dict['COMPLETE'] - passes
+            fields = ['State', 'Amount', 'Percent', 'PASSED', 'FAILED']
+
+        for key, value in summary_dict.items():
+            #  Build the rows for drawtables.
+
+            #  Determine Color.
+            if key in {'ERROR', 'FAILED', 'TIMEOUT'}:
```
Paul-Ferrell, line 371: There are a bunch of states that contain 'ERROR', 'FAILED' and 'TIMEOUT'. All of them should be RED.

lib/pavilion/plugins/commands/status.py
```diff
@@ -321,3 +328,90 @@ def display_history(self, pav_cfg, args):
...
+                color = 'GREEN'
+            elif key in 'SKIPPED':
+                color = 'YELLOW'
+            elif key in {'RUNNING', 'SCHEDULED', 'PREPPING_RUN'}:
+                color = 'CYAN'
+            else:
+                color = 'WHITE'  # Unknown state.
+
+            # Populating rows...
+            if key in 'COMPLETE':  # only time we need to populate pass/fail
```
Paul-Ferrell, line 383: if the key is 'C' it will match this as true...  It probably works, because the only key that would be in 'COMPLETE' is the word 'COMPLETE' itself.

lib/pavilion/plugins/commands/status.py
```diff
@@ -321,3 +328,90 @@ def display_history(self, pav_cfg, args):
...
+        if 'COMPLETE' in summary_dict.keys():
+            fails = summary_dict['COMPLETE'] - passes
+            fields = ['State', 'Amount', 'Percent', 'PASSED', 'FAILED']
+
+        for key, value in summary_dict.items():
+            #  Build the rows for drawtables.
+
+            #  Determine Color.
+            if key in {'ERROR', 'FAILED', 'TIMEOUT'}:
+                color = 'RED'
```
Paul-Ferrell, line 372: All the colors are available as constants in the output module. Use them.

```python
color = output.GREEN
```

Issue #250, opened 2020-05-29
Allow for non-relocatable builds


nicholas-sly:
Provide an option that allows the test to be built in the final location rather than in a temporary directory for tests that aren't amenable to relocation.  A new config option under the build section of `relocatable: true` that is `true` by default and allows for `false` would allow for the specification.  

A possible alternative would be to create the final directory before starting the build and allow it to be referenced using a variable so builds that allow you to specify a prefix can use that.  

Either way, unless LAP has some special mechanism to allow the builds to be relocatable that I haven't found yet (already sent an email inquiring about this), this will enable building and running LAP/FLAG under Pavilion.

Paul-Ferrell:
This was fixed in another merge request recently.

PR #251, opened 2020-06-01
WIP: Add in_place build functionality... hopefully.

nicholas-sly:
Trying to add the ability to specify that a test should be built in the final location instead of a temporary file and moved to the final location.  This has become necessary when integrating tests that integrate the build location into the test in a way that makes it non-portable.  Outlined in #250 .

The current implementation idea is just to add a key under the build section of a test 'in_place' that modifies the behavior of the builder class to not add `.tmp` to the end of the build directory (likewise skipping the step to move it to the final location).

### DIFF
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -411,7 +411,10 @@ class TestBuilder:
                     self.tracker.update(
                         state=STATES.BUILDING,
                         note="Starting build {}.".format(self.name))
-                    build_dir = self.path.with_suffix('.tmp')
+                    if not self._config.get('in_place'):
+                        build_dir = self.path.with_suffix('.tmp')
+                    else:
+                        build_dir = self.path
 
                     # Attempt to perform the actual build, this shouldn't
                     # raise an exception unless something goes terribly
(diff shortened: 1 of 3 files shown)


Paul-Ferrell:
You have to have some sort of locking in place, and the final determination that a build is ready has to be through some sort of atomic mechanism.

Right now:
  - The build checks for the existence of the build directory. If it doesn't exist, it knows it _MAY_ have to perform a build. 
  - Then it gets a lock on the tmp build directory. 
  - If it has the lock, it checks again for the final build directory's existence. If it's there, then we're done.
  - If the build directory didn't exist, create the tmp build directory and build. 
  - Then, do an atomic 'rename'/'mv' operation to move the tmp build to the final build location.
  - Finally, release the lock.

This allows for builds to locked across hosts as long as atomic move/rename and atomic file creation (used in the locking system) are supported on the filesystem.

My plan to fix this:

 - The final indication of a build's completion is no longer the directory itself, it would be a '<build_hash>.complete' (build.complete) file. 
 - So check for the existence of that file (and the build itself. if it's missing we have other problems.)
 - Then get a lock on the actual build directory.
 - With the lock, perform the build (if the build directory and build.complete don't already exist).
 - Make the build.complete file (atomically).
 - Release the lock.

Paul-Ferrell:
The problem with the above is that it adds extra cases:
 1. The build dir doesn't exist but build.complete does.
     - We should probably issue a warning, get the lock, delete build.complete, release the lock, proceed cleanly.
 2. The build dir exists, the build.complete file doesn't, and the lock can be obtained (locks have an expiration date).
     - The problem is that another test could conceivably be using that build, and there's no way to safely check for that. 
     - In this case, we should deprecate the build, and then perform a rebuild.
     -

PR #253, opened 2020-06-02
Fixed issue with evaluation ordering.

Paul-Ferrell:
I originally hoped implicit ordering on dictionaries could take care of this, but it doesn't. I've changed things so that order doesn't matter now.


Paul-Ferrell:
The config path normalization ticket should take care of the issues you were having.

PR #254, opened 2020-06-02
Fixed normalization issue.

Paul-Ferrell:


Paul-Ferrell:
Make sure this is unique?

PR #258, opened 2020-06-08
Significant Documentation Improvements

Paul-Ferrell:


docs/advanced.rst
```diff
@@ -265,8 +256,12 @@ environment (or lmod) modules.
...
 ~~~~~~~~~~~~~~~~~~~~~
 
-You can set/unset environment variables in your test configs, and use
-them in your scripts.
+*Full Docs:* :ref:`tests.env.variables`
+
+You can set environment variables in your test scripts using the
+'env' section under both 'run' and 'build'. This will cause the variables to
+be exported within the generated run or build script, where they can be used
+by commands run as part of that script.
```
francinelapid, line 264:
```suggestion
'env' section under both 'run' and 'build'. This will cause the variables to
be exported within the generated run or build script, where they can be used
by commands run as part of that script. You may only reference environment
variables in the 'cmds' subsection section of the 'run' and 'build' sections,
as they are the only sections that are converted into shell scripts. 
```

docs/test_run_lifecycle.rst
```diff
@@ -0,0 +1,460 @@
...
+For every test given as part of the run command, Pavilion will find the
+relevant test files and generate a raw config structure.
+
+Finding Suite, Host, and Mode Configs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Pavilion can be configured to look in multiple places for test configs
+(see :ref:`config.config_dirs`), and uses the first matching file found. In
+case, we need a suite config named ``tests/mytests.yaml`` and a host config
+named ``host/tester.yaml``. Pavilion will load the yaml from each of these
```
francinelapid, line 113:
```suggestion
named ``hosts/tester.yaml``. Pavilion will load the yaml from each of these
```

docs/tests/run.rst
```diff
@@ -112,207 +103,9 @@ The list of commands to perform when running the test.
...
+^^^^^^^
 
-Lastly, the test run is set as complete, regardless of whether it
-passed, failed, or encountered an error. Note that this is separate from
-the status file; a file named 'RUN\_COMPLETE' is created in the test run
-directory. The file contains only a timestamp of when the run officially
-ended. Various commands can use this as an easy way to differentiate
-complete tests from those that may still be running.
+By default test runs timeout if they don't produce output within a certain
+amount of time. This setting allows you to extend that time arbitrarily,
```
francinelapid, line 110: You specified default build timeout, I think you should specify it here also. 
```suggestion
By default test runs timeout if they don't produce output within a 30 seconds. 
This setting allows you to extend that time arbitrarily,
```

lib/pavilion/plugins/results/command.py
```diff
@@ -1,3 +1,5 @@
+"""Execute a command and get it's output or return value."""
```
francinelapid, line 1:
```suggestion
"""Execute a command and get its output or return value."""
```

PR #261, opened 2020-06-11
Show sub command

### LINKED ISSUE #182: Add show command to list host and mode files.
Should have a cfg sub-command (or --cfg option) to show the full config.
Should have a vars sub-command (or --vars option) to show just the defined variables.

kjeverson:
Closes #182


Paul-Ferrell:
Do you really think that adding this stuff as additional columns to the table, and displaying the full configs all at once for all host files, is the way to go? I... don't.

Take a step back, figure out a reasonable, usable way to approach this problem, and try again. I can dictate a solution, but I need to not have to. 

You can do better.

lib/pavilion/plugins/commands/show.py
```diff
@@ -385,65 +412,117 @@ def _functions_cmd(self, _, args):
...
+                    if file.suffix == '.yaml' and file.is_file():
+                        host_id = file.stem
+                        host_path = file
+                        with file.open() as config_file:
+                            config_data = yc_yaml.load(config_file)
+                        try:
+                            host_vars = list(config_data['variables']
+                                             .keys())
+                        except (KeyError, TypeError) as err:
+                            host_vars = []
```
Paul-Ferrell, line 440: You should load these with the test config loader, not as raw yaml.

```python
from pavilion.test_config import file_format

config = file_format.TestConfigLoader().load(myfile)
```

This will guarantee that all the config fields exist in the resulting structure, though they may be empty. So the 'variables' key will always exist, for instance, and will always be a dictionary.

Paul-Ferrell: Ok, so the --vars output still shows what variables are available. 
  - But is a list of their names really readable? 
  - What are their defined sub-keys (if any)
  - What are their values? 
  - How do they compare across host files?

There's lots of stuff that would be useful to do here, and it's going to require a completely separate table from the standard one.

lib/pavilion/plugins/commands/show.py
```diff
@@ -385,65 +412,117 @@ def _functions_cmd(self, _, args):
...
-                    })
+            output.draw_table(
+                self.outfile,
+                field_info={},
+                fields=col_names,
+                rows=hosts
+            )
+
+        else:
+
```
Paul-Ferrell, line 456: Instead of a giant if/else, break these out into different methods.

lib/pavilion/plugins/commands/show.py
```diff
@@ -385,65 +412,117 @@ def _functions_cmd(self, _, args):
...
+                return errno.EINVAL
 
-        output.draw_table(
-            self.outfile,
-            field_info={},
-            fields=col_names,
-            rows=hosts
-        )
 
     @show_cmd('mode')
```
Paul-Ferrell, line 470: Wait a second, isn't 99.9% of this code the same as the above?

You should make this stuff generic.

lib/pavilion/plugins/commands/show.py
```diff
@@ -385,65 +412,117 @@ def _functions_cmd(self, _, args):
...
 
-            for file in os.listdir(path.as_posix()):
+                for file in os.listdir(path.as_posix()):
+
+                    file = path / file
+                    if file.suffix == '.yaml' and file.is_file():
+                        host_id = file.stem
+                        host_path = file
+                        with file.open() as config_file:
+                            config_data = yc_yaml.load(config_file)
```
Paul-Ferrell, line 435: Move this logic into a method of the test_config.resolver.Resolver class that finds and loads all of the host or mode files. There's already an equivalent there for test configs.

lib/pavilion/plugins/commands/show.py
```diff
@@ -780,3 +859,19 @@ def _tests_cmd(self, pav_cfg, args):
     def _test_config_cmd(self, *_):
         """Show the basic test config format."""
         file_format.TestConfigLoader().dump(self.outfile)
+
+    def get_config(self, pav_cfg, name, directory):
```
Paul-Ferrell, line 863: There's already a 'find' method in test_config.resolver that handles this.

Paul-Ferrell: Though it doesn't load the config, just finds it.

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
+
+        name = config
+        path = file
+        config = config_data
+
+        for var in config['variables'].keys():
+            var_out = var
+            index = 0
+            for subvar in config['variables'][var]:
+                if type(subvar) is yaml_config.elements.ConfigDict:
```
Paul-Ferrell, line 436: You want `isinstance(dict)`, generally.

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
+                    index += 1
+                else:
+                    data.append({
+                        'Variables': var_out,
+                        ' ': index,
+                        'Type': type(subvar).__name__,
+                        'Value': subvar,
+                    })
+                    var_out = ''
+                    index += 1
```
Paul-Ferrell, line 459: I mostly like what you're doing here, but:
1. Give the index column a name. Even even if it makes the column a little fat.
2. You don't have to capitalize the field names, that happens automatically.
3. I don't like the type column, it's confusing. Every variable has an index and is a 'list' of strings or dict of strings.
4. This doesn't give very readable output for multi-valued variables. More on that below.

Paul-Ferrell: Or right here. 

For each multi-valued variable, make a separate table for each. You already have the rows, just annotate them with an index column. Add a title that says what the variable name is, etc.

This will be ugly for very large sets of variables. Good. Jen likes to do that, and it's horrendous.

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
+        col_names = ['Variables', ' ', 'Type', 'Value']
+
+        name = config
+        path = file
+        config = config_data
+
+        for var in config['variables'].keys():
+            var_out = var
+            index = 0
+            for subvar in config['variables'][var]:
```
Paul-Ferrell, line 435: You spend so much space tracking the index; Just loop over it.
```python
for idx in range(len(config['variables'][var])):
  subvar = config['variables'][var][idx]
```

Then you just have it. No need to increment it at the right places, and you can and should start doing things based on it. For example:
```python
                        data.append({
                            'variables': var_out if idx == 0 else '',
                            ' idx': idx,
                            'Value': subvar,
                        })
```

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
+                        'Value': subvar,
+                    })
+                    var_out = ''
+                    index += 1
 
-        hosts = []
+        output.fprint("Showing Variables for {} {} config at {}: "
+                      .format(name, conf_type.strip('s'), path))
+
+        if data:
```
Paul-Ferrell, line 463: If there isn't anything. Print a message that says so.

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
+            output.draw_table(
+                self.outfile,
+                field_info={},
+                fields=col_names,
+                rows=data
+            )
+
+    def show_table(self, pav_cfg, args, conf_type):
```
Paul-Ferrell, line 471: 'show_table' of what?

Paul-Ferrell: Maybe `list_sub_configs`?

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
-                file = path / file
-                if file.suffix == '.yaml' and file.is_file():
-                    mode_id = file.stem
-                    mode_path = file
-                    modes.append({
-                        'Name': mode_id,
-                        'Path': mode_path
-                    })
+        if not args.show_cmd.endswith('s'):
+            args.show_cmd = args.show_cmd + 's'
```
Paul-Ferrell, line 518: Why not just set show_cmd to 'hosts' or 'modes' in the calling function?

lib/pavilion/test_config/resolver.py
```diff
@@ -208,6 +209,44 @@ def default(val, dval):
...
+            for file in os.listdir(path.as_posix()):
+
+                file = path / file
+                if file.suffix == '.yaml' and file.is_file():
+                    name = file.stem
+                    configs[name] = {}
+
+                    full_path = file
+                    with file.open() as config_file:
+                        config = file_format.TestConfigLoader().load(config_file)
```
Paul-Ferrell, line 243: You need to catch errors here, and add the message as a separate attribute.

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
+                                                                 config)
+        with file.open() as config_file:
+            config_data = file_format.TestConfigLoader().load_raw(config_file)
 
-            if not (path.exists() and path.is_dir()):
-                continue
+        if config_data is not None:
+            output.fprint("{} config for {} found "
+                          "at:{}".format(conf_type.strip('s'), config,
+                                         str(file)), file=self.outfile)
```
Paul-Ferrell, line 507: The table has this info, you don't need to repeat it here.

lib/pavilion/plugins/commands/show.py
```diff
@@ -390,69 +415,126 @@ def _functions_cmd(self, _, args):
...
+                field_info={},
+                fields=col_names,
+                rows=data
+            )
+
+    def show_table(self, pav_cfg, args, conf_type):
+
+        configs = resolver.TestConfigResolver(pav_cfg).find_all_configs(conf_type)
+
+        data = []
```
Paul-Ferrell, line 475: A one column table is kind of sad. Might as well drop verbose and always print the path.

Instead, you do need to add an option (like with `pav show tests` to print any errors encountered when loading the config.

Paul-Ferrell (commented):
Fix the stuff I mentioned in chat.

Paul-Ferrell (dismissed):

Paul-Ferrell (approved):

PR #263, opened 2020-06-12
add time flag + searchable series,

CalvinDSeamons:
This is not ready for a review.  I cant find the "work in progress label", it's Friday...


lib/pavilion/plugins/commands/status.py
```diff
@@ -139,6 +140,65 @@ def get_tests(pav_cfg, args, errfile):
...
+        time_amount = time_amount * 168
+    elif time_unit == 'month' or time_unit == 'months':
+        time_amount = time_amount * 720
+    elif time_unit == 'year' or time_unit == 'years':
+        time_amount = time_amount * 8760
+    else:
+        output.fprint("Invalid time unit", color=output.RED)
+
+    search_date = datetime.today() - timedelta(hours=time_amount)
+
```
Paul-Ferrell, line 177: Make a limit of None do 'no limit'

CalvinDSeamons:
As for moving code into utils for reduce redundancy, right now `--time` can be accessed by `--series` and `pav status`. Would you want a universal utils command that returns statuses based on datetime? For display_history display_series and display_summary those are all tied to `status` and should stay in `status.py`.

lib/pavilion/plugins/commands/status.py
```diff
@@ -139,6 +140,83 @@ def get_tests(pav_cfg, args, errfile):
...
+    else:
+        output.fprint("Invalid time unit, --time only accepts "
+                      "second(s), minute(s), hour(s), day(s), "
+                      "week(s), month(s), and year(s).",
+                      file=errfile, color=output.RED)
+        return []  # No tests to return.
+
+    try:
+        search_date = datetime.today() - timedelta(hours=time_amount)
+    except OverflowError:
```
Paul-Ferrell, line 187: This isn't quite true. Datetime can't go back before year 1 AD.

lib/pavilion/plugins/commands/status.py
```diff
@@ -139,6 +140,83 @@ def get_tests(pav_cfg, args, errfile):
...
+    :returns: List of dictionary objects"""
+    # Converting time arguments into valid datetime object.
+    try:
+        time_amount = int(args.time[0])
+        time_unit = args.time[1]
+    except ValueError:
+        output.fprint("Invalid Format. Use the from: $AMOUNT $UNIT e.g."
+                      " '5 hours'", file=errfile, color=output.RED)
+        return []  # No tests to return.
+
```
Paul-Ferrell, line 160: Put this in a function under utils.py

lib/pavilion/plugins/commands/status.py
```diff
@@ -255,20 +333,31 @@ def _setup_arguments(self, parser):
             '--history', type=int,
             help="Shows the full status history of a job."
         )
+        parser.add_argument(
```
Paul-Ferrell, line 336: Make this a sub-command, not a regular argument.

`pav status series <args>`. See the show command for an example.

lib/pavilion/plugins/commands/status.py
```diff
@@ -255,20 +333,31 @@ def _setup_arguments(self, parser):
...
+            '--series', type=str, default=False, nargs=2,
+            help='Displays test series based on user specifications.'
+        )
         parser.add_argument(
             '-s', '--summary', default=False, action='store_true',
             help='Summary will display a summed version of what'
                  'state were observed in pav status. '
         )
```
Paul-Ferrell, line 344: What happened to filtering by test name, user, and host (sys_name)?

lib/pavilion/plugins/commands/status.py
```diff
@@ -255,20 +333,31 @@ def _setup_arguments(self, parser):
...
+        parser.add_argument(
+            '-t', '--time', type=str, default=False, nargs=2,
+            help='Retrieves tests up to the given date or time.'
+        )
 
     def run(self, pav_cfg, args):
         """Gathers and prints the statuses from the specified test runs and/or
         series."""
         try:
-            if not args.all:
```
Paul-Ferrell, line 268: This all needs to be simplified.

lib/pavilion/plugins/commands/status.py
Paul-Ferrell: It doesn't need to return tests, just a cutoff datetime to use in a filter.

lib/pavilion/plugins/commands/status.py
CalvinDSeamons: I already catch that.

lib/pavilion/plugins/commands/status.py
CalvinDSeamons: You told me to just do '--user' and '--sys_name`. I can add those if you want.

CalvinDSeamons:
Too much has changed to even attempt to reuse this branch. Series will also be it's own command now.

PR #265, opened 2020-06-16
Pass shared_group to script composer.

paulbry:
I ran into issues on a system where the group did not match the user. I believe specifying the `shared_group` in the config should have corrected this but I ended up seeing"

```
Unknown error running command run: Group cschisd not found on this machine..
Traceback (most recent call last):
    ...
    raise ScriptComposerError(error)
pavilion.scriptcomposer.ScriptComposerError: Group cschisd not found on this machine.
```

This PR seeks to address this.

From what I could tell the use of `pav_cfg.shared_group` should be acceptable since a `None` value found in an undefined config should be accounted for. Though please let me know if I'm missing something.


Paul-Ferrell:
I've run into this. I've got a broader fix lined up, but it's still in progress. 

Thanks!

Issue #65, opened 2019-11-27
Add a 'version' section for test configs


nicholas-sly:
Add a version key to the base subtest config format.  Have this also included as a default result key.  The default value can be something like '0.1' or '1.0'.

Paul-Ferrell:
Three version actually
 - Test version (arbitrary, save in results)
 - Min Pavilion Version
 - Current Pavilion version saved in results

Paul-Ferrell:
- Make the pavilion version a pavilion variable.

Paul-Ferrell:
- Min Pav Version should be a range.

PR #268, opened 2020-06-17
Test config version

kjeverson:
Addresses issue #65 

Test Configs now support version sections that looks like:
```yaml
Test:
    test_version: 1.2.3
    compatible_pav_versions: 2.1.0-2.2.8
    Build:
    .
    .
    .
```

Compatible_pav_version currently supports version to be entered as follows:
- Single version, ex: `1.2.3`, `1.2`
- Version range, ex: `1.2.3-1.4.5`, `1.2-1.4.5`, etc.


lib/pavilion/result/base.py
```diff
@@ -28,10 +28,12 @@ def get_sched_keys(test):
 
     return sched_keys
 
-
 BASE_RESULTS = {
     'name': lambda test: test.name,
     'id': lambda test: test.id,
+    'test_version': lambda test: test.test_version,
+    'pav_version': lambda test: test.var_man['pav.version'],
+    'compatible_pav_versions': lambda test: test.compatible_pav_versions,
```
Paul-Ferrell, line 36: I don't think this one needs to be in the results. The other two do though.

lib/pavilion/test_config/file_format.py
```diff
@@ -16,10 +16,10 @@ class TestConfigError(ValueError):
 TEST_NAME_RE_STR = r'^[a-zA-Z_][a-zA-Z0-9_-]*$'
 TEST_NAME_RE = re.compile(TEST_NAME_RE_STR)
 KEY_NAME_RE = re.compile(r'^[a-zA-Z][a-zA-Z0-9_-]*$')
+VERS_RANGE_RE = re.compile(r'\d+\.\d+\.\d+-\d+\.\d+\.\d+')
```
Paul-Ferrell, line 19: I don't think you should force a range. Let's accept things like the following:

2.1.0
2.1.*
2.1-2.2.3

test/tests/resolver_tests.py
```diff
@@ -466,3 +471,64 @@ def test_env_order(self):
...
+            'version_compatible.three': {
+                'test_version': '1.0',
+                'pav_version': pav_version,
+                'compatible_pav_versions': 'none'
+            }
+        }
+
+        # Ensures Version information gets populated correclty even with empty
+        # version section in test config
+        run_cmd = commands.get_command(args.command_name)
```
Paul-Ferrell, line 509: There's are useful .silence() and .clear_output() methods on all commands that do this setup now.

lib/pavilion/result/base.py
kjeverson: Which one? All three are highlighted.

Paul-Ferrell: 'compatable_pav_versions'. It looks different when you comment on multiple lines, but it's definitely confusing.

test/tests/resolver_tests.py
```diff
@@ -466,3 +471,63 @@ def test_env_order(self):
...
+            },
+            'version_compatible.two': {
+                'test_version': 'beta',
+                'pav_version': pav_version,
+                'compatible_pav_versions': '1.2.3-5.4.9'
+            },
+            'version_compatible.three': {
+                'test_version': '0.0',
+                'pav_version': pav_version,
+                'compatible_pav_versions': 'all'
```
Paul-Ferrell, line 503: Multi-line comment.

test/tests/resolver_tests.py
```diff
@@ -466,3 +471,63 @@ def test_env_order(self):
...
+        expected_results = {
+            'version_compatible.one': {
+                'test_version': '1.2.3',
+                'pav_version': pav_version,
+                'compatible_pav_versions': 'all'
+            },
+            'version_compatible.two': {
+                'test_version': 'beta',
+                'pav_version': pav_version,
+                'compatible_pav_versions': '1.2.3-5.4.9'
```
Paul-Ferrell, line 498: Single line comment.

lib/pavilion/test_config/resolver.py
```diff
@@ -440,6 +449,106 @@ def load_raw_configs(self, tests, host, modes):
...
+
+        # User may provide compatible_pav:versions in the following formats:
+        # Single Version, EX: 1.2.3
+        # Wild Card Version, EX: 1.2 or 1.2.*
+        #   Note: These have the same behavior
+        # Range of Versions, EX: 1.2.3-1.4.5, 1.2-3.4, 1.2-1.2.4, etc.
+
+        ABBRV_RE = re.compile(r'^\d+\.\d+$')
+        FULL_RE = re.compile(r'^\d+\.\d+\.\d+$')
+        WC_RE = re.compile(r'^\d+\.\d+\.\*$')
```
Paul-Ferrell, line 464: you only need one regex.

Valid patterns ('n' is a sequence of digits):
- n
- n.*
- n.n
- n.n.*
- n.n.n

Groups and `|` patterns are your friend:
`r'a(bc|de)?'` matches `a`, `abc`, or `ade`

lib/pavilion/test_config/resolver.py
```diff
@@ -440,6 +449,106 @@ def load_raw_configs(self, tests, host, modes):
...
+                              "\ncompatible_pav_versions '{}' is not a "
+                              "valid version format."
+                             .format(test, test_suite, compatible_versions))
+                        raise TestConfigError(err)
+
+                    if not pav_version == version:
+                        compatible_errors += err.format(test,
+                                                        PavVars.version(self),
+                                                        compatible_versions)
+
```
Paul-Ferrell, line 546: Regardless of the multiple regular expressions, this should still be a lot simpler. 
 - Figure out a discrete set of steps to follow. 
 - The output from each step should be normalized to minimize special cases in the next step.
 - Break a step out into a function only if it's complicated enough that you can't tell what the steps' code does at
   a glance, if the function will be re-used, or to further divide the problem into sub-steps. 

1. get_min_str('1.2.*-3.4') -> '1.2.*'
2. get_max_str('1.2.*-3.4') -> '3.4'
3. calc_min(min_str) -> (1, 2, 0)
4. calc_max(max_str) -> (3, 4, None)
5. check_version(version, min_version, max_version) -> True|False

Paul-Ferrell: Draw table is both a good and bad example. It's good in that it follows discrete steps to get to the end result. It's bad in that each of the larger steps should be broken out into separate functions, which is mostly a result of the organic way in which the function grew over time.

lib/pavilion/test_config/resolver.py
```diff
@@ -455,6 +467,102 @@ def load_raw_configs(self, tests, host, modes):
...
+        else:
+            return versions
+
+    def calc_min(self, min_str):
+        """Ensures version was provided in the correct format, and returns the
+        lower bound as a list of digits."""
+
+        result = TEST_VERS_RE.match(min_str)
+        if result is not None:
+            min_version = min_str.split(".")
```
Paul-Ferrell, line 493: Let's just drop the idea of '*', being a thing.  It will simplify everything.

lib/pavilion/test_config/resolver.py
```diff
@@ -393,6 +397,14 @@ def load_raw_configs(self, tests, host, modes):
                         "Test suite '{}' raised a type error, but that "
                         "should never happen. {}".format(test_suite_path, err))
 
+                # Check test compatibility with current pavilion version
+                comp_error = self.check_suite_version_compatibility(test_suite_cfg)
```
Paul-Ferrell, line 401: You should only check this if validation fails on the config. The best place to do that is during validation in 'resolve_inheritance', where it can be done on a test-by-test basis.

lib/pavilion/test_config/resolver.py
```diff
@@ -393,6 +397,14 @@ def load_raw_configs(self, tests, host, modes):
...
+
+                if comp_error:
+                    raise TestConfigError(
+                        "Test suite '{}' has incompatibility issues: {}"
+                        .format(test_suite, comp_error))
+
```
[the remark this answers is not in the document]
kjeverson: That's exactly why I did it this way, user will see all incompatibility issues at once so they only need to go and edit the test suite once. Instead of only seeing the firs occurrence, changing, and seeing another issue. I can change it, but feel like it's better to display all issues at once...

lib/pavilion/test_config/file_format.py
```diff
@@ -260,6 +260,17 @@ class TestConfigLoader(yc.YamlConfigLoader):
                       "the Pavilion variable matches one or more of the "
                       " values."
         ),
+        yc.StrElem(
+            'compatible_pav_versions', default='',
+            help_text="Specify compatibile pavilion versions for this "
+                      "specific test. Can be represented as a single "
+                      "version, ex: 1, 1.2, 1.2.3, a wildcard, ex: 1.*, "
+                      "1.2.*, or a range, ex: 1.2-1.3.4, etc."
```
Paul-Ferrell, line 268: You've dropped the need for '*', but not from the docs.

Issue #260, opened 2020-06-10
Per sys_name tracking of last series ran


Paul-Ferrell:
Pavilion currently tracks the last series a user ran, but that's pretty useless if you're running pavilion a across multiple systems. 

 - Store a json file that contains an entry for each sys_name for a given user.
 - Use pavilion.lockfile locks to make sure this is atomic.

PR #272, opened 2020-06-17
Sys name series tracking

kjeverson:
Addresses issue #260


lib/pavilion/series.py
```diff
@@ -155,6 +161,38 @@ def _save_series_id(self):
             self._logger.warning("Could not save series id to '%s'",
                                  last_series_fn)
 
+    def _sys_name_series_tracking(self):
```
Paul-Ferrell, line 164: You should be replacing the old way of tracking series for users, not adding a separate way.

lib/pavilion/series.py
```diff
@@ -155,6 +161,38 @@ def _save_series_id(self):
...
+        on a per system basis."""
+
+        sys_vars = system_variables.get_vars(True)
+        sys_name = sys_vars['sys_name']
+
+        json_file = self.pav_cfg.working_dir/'users'
+        json_file /= '{}.json'.format(utils.get_login())
+
+        lockfile_path = json_file.parent/(json_file.name + '.lock')
+        lockfile = LockFile(lockfile_path)
```
Paul-Ferrell, line 175: I didn't lock in the original function because it wasn't necessary. In this case though it is only necessary to prevent loss of data, not file conflicts. Because you're writing a new file every time, you never have to worry about corrupting the read of some other process. You may have a situation like this though:

- proc A reads
- proc B reads
- proc A writes
- proc B writes

In which case, the final file will be missing the data changed by proc A. 

You may already know all this, but I thought it would be good to point out.

lib/pavilion/series.py
```diff
@@ -140,36 +145,55 @@ def from_id(cls, pav_cfg, id_):
...
-            with last_series_fn.open('w') as last_series_file:
-                last_series_file.write(self.id)
-        except (IOError, OSError):
-            # It's ok if we can't write this file.
-            self._logger.warning("Could not save series id to '%s'",
-                                 last_series_fn)
+        json_file = self.pav_cfg.working_dir/'users'
+        json_file /= '{}.json'.format(utils.get_login())
+
+        lockfile_path = json_file.parent/(json_file.name + '.lock')
```
Paul-Ferrell, line 157:
```python
lockfile_path = json_file.with_suffix('.lock')
```

lib/pavilion/series.py
```diff
@@ -140,36 +145,55 @@ def from_id(cls, pav_cfg, id_):
...
+        lockfile = LockFile(lockfile_path)
+
+        with lockfile:
```
Paul-Ferrell, line 160:
```python
with LockFile(lockfile_path):
    ...
```

Paul-Ferrell:
Fixes #260

Issue #271, opened 2020-06-17
Add some config fields, change how source downloading works.


Paul-Ferrell:
Add: 
 - maintainer
   - name
   - email

Change:
 - build.source_location to `build.source`
 - remove `source_name`
 - add `source_url`
 - add `download=never | missing | latest`

PR #273, opened 2020-06-18
Added new config elements

Paul-Ferrell:
- Changed how file downloads work.

### DIFF
--- a/lib/pavilion/test_config/file_format.py
+++ b/lib/pavilion/test_config/file_format.py
@@ -306,8 +306,10 @@ expected to be added to by various plugins.
                     'source_path',
                     help_text="Path to the test source. It may be a directory, "
                               "compressed file, compressed or "
-                              "uncompressed archive (zip/tar). For relative "
-                              "paths Pavilion looks in the test_src directory "
+                              "uncompressed archive (zip/tar), and is handled "
+                              "according to the internal (file-magic) type. "
+                              "For relative paths Pavilion looks in the "
+                              "test_src directory "
                               "within all known config directories. If this"
                               "is left blank, Pavilion will always assume "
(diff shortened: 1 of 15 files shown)


Paul-Ferrell:
Fixes #271

lib/pavilion/test_config/file_format.py
```diff
@@ -293,22 +303,31 @@ class TestConfigLoader(yc.YamlConfigLoader):
...
+                              "uncompressed archive (zip/tar). For relative "
+                              "paths Pavilion looks in the test_src directory "
+                              "within all known config directories. If this"
+                              "is left blank, Pavilion will always assume "
+                              "there is no source to build."),
+                yc.StrElem(
+                    'source_url',
+                    help_text='Where to find the source on the internet. By '
+                              'default, Pavilion will try to download the '
+                              'source from the given URL the source file '
```
nicholas-sly, line 318: '...the given URL if the source...'

lib/pavilion/test_run.py
```diff
@@ -234,20 +243,10 @@ def __init__(self, pav_cfg, config,
...
-                                    "'source_download_name is set without a "
-                                    "'source_location'")
-                    raise TestConfigError(msg)
-        except KeyError:
-            # this is mostly for unit tests that create test configs without a
-            # build section at all
-            pass
+        if (build_config.get('source_path') is None and
+                build_config.get('source_url') is not None):
+            raise TestConfigError(
```
nicholas-sly, line 248: This is an interesting case.  I guess we're requiring that the test creator know the form that the test will have after being downloaded.  In the case of 'source_download: latest', could we not just determine the 'source_path' based on what we get from downloading 'source_url'?

Issue #275, opened 2020-06-19
'only_if' and 'not_if' should keep the build from happening as well as the run


nicholas-sly:
The 'only_if' and 'not_if' need to stop the build from happening as well as the run.  Trying to exclude a test from running on a toss system that uses cray-specific modules to build results in a failure.

Paul-Ferrell:
Fixed in 2.3 release

PR #277, opened 2020-06-22
pushed up the point of _get_skip to call before build

CalvinDSeamons:
Nick pointed out a serious problem of tests that should skip but are failing instead. This is due to commands in the build section of the tests running before it evaluates the skip. For example the build section might say load some cray library. This will fail on TOSS before skip is called. 

I moved where skip is being called to the earliest part of the `__init__` in test_run.py. Hopefully this will work and if not some serious rework is required, yikes...

### DIFF
--- a/lib/pavilion/plugins/commands/run.py
+++ b/lib/pavilion/plugins/commands/run.py
@@ -149,11 +149,7 @@ class RunCommand(commands.Command):
             self._complete_tests(all_tests)
             return res
 
-        runnable_test = []
-        for test in all_tests:
-            if not test.skipped:
-                runnable_test.append(test)
-        all_tests = runnable_test
+        all_tests = [test for test in all_tests if not test.skipped]
 
         res = self.build_local(
             tests=all_tests,
--- a/lib/pavilion/plugins/commands/run.py
+++ b/lib/pavilion/plugins/commands/run.py
@@ -149,6 +149,12 @@ class RunCommand(commands.Command):
             self._complete_tests(all_tests)
             return res
 
+        runnable_tests = []
+        for test in all_tests:
+            if not test.skipped:
+                runnable_tests.append(test)
+        all_tests = runnable_tests
+
         res = self.build_local(
             tests=all_tests,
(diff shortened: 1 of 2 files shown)


CalvinDSeamons:
This addresses issue #275, maybe...

lib/pavilion/test_run.py
```diff
@@ -292,7 +292,9 @@ def __init__(self, pav_cfg, config,
         self._results = None
         self._created = None
 
-        self.skipped = self._get_skipped()
+        self.skipped = self._get_skipped()  # eval skip.
+        if self.skipped:
```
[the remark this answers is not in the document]
CalvinDSeamons: Sorry the `if self.skipped: return None` was redundant in trying to fix the compounding skip states being added. I was able to fix that in the `_get_skipped` function.

lib/pavilion/plugins/commands/run.py
```diff
@@ -149,6 +149,12 @@ def run(self, pav_cfg, args):
             self._complete_tests(all_tests)
             return res
 
+        runnable_test = []
+        for test in all_tests:
+            if not test.skipped:
+                runnable_test.append(test)
+        all_tests = runnable_test
```
Paul-Ferrell, line 156:
```python
all_tests = [test for test in all_tests if not test.skipped]
```

lib/pavilion/test_run.py
```diff
@@ -988,8 +988,13 @@ def __repr__(self):
         return "TestRun({s.name}-{s.id})".format(s=self)
 
     def _get_skipped(self):
-        skip_reason_list = self._evaluate_skip_conditions()
-        matches = " ".join(skip_reason_list)
+        """Kicks off assessing if current test is skipped."""
+        if self.status.current().state == 'SKIPPED':
```
Paul-Ferrell, line 992: You shouldn't rely on status like this. Add 'skipped' as a test attribute (there's a new 'attributes' file).

CalvinDSeamons: See my comment on mattermost.

PR #283, opened 2020-07-07
Moved datetime resolver and cleaned up pav status

CalvinDSeamons:


CalvinDSeamons:
First wack and trying to clean the status command up. I condensed stuff into a new method `trim_tests` and moved the datetime resolver to utils. Right now the hierarchy for the flags is as follows. `--time, --limit` are kinda tied and `--all` is least important.
`pav status --all --time 2 seconds` // all tests in last 2 seconds (if nothing ran in 2s it returns nothing, not all test...)
`pav status --all --limit 3` // returns 3 tests
`pav status --time 500 years --limit 5` // looks at all tests in the last 500 years but only return 5 most recent
`pav status --all` // returns all tests
`pav status --all --time 5 months --limit 2` //looks at all tests returns 5 months worth but only show 2 with the limit.
`pav status` // default limit of 10, returns 10 most recent tests.

lib/pavilion/plugins/commands/status.py
```diff
@@ -139,6 +112,57 @@ def get_tests(pav_cfg, args, errfile):
...
+    """Consolidates and filters test_status list based on
+    the limiting argument flags under 'pav status'. """
+
+    test_statuses = []
+    test_obj_list = []
+
+    if args.all and not args.limit:
+        tests = test_run.get_latest_tests(pav_cfg)
+    else:
+        tests = test_run.get_latest_tests(pav_cfg, args.limit)
```
Paul-Ferrell, line 125: You're limiting test by time twice. Once by get_latest_tests standards, and once by your standards. The first filtering may return different tests than you would expect by your standards. 

According to pycharm, the only usage of get_latest_tests is in the status command. 
Here's what I think you should do: 
  - Get rid of the 'get_latest_tests' function (entirely)
  - Use dir_db.select instead. 

You'll need to dynamically generate a filter function. That will be educational.

Paul-Ferrell:
Addressed in other merge requests.

PR #288, opened 2020-07-15
Updated release number.

Paul-Ferrell:


Paul-Ferrell:
Can you guys both check for anything I might have missed in the release notes?

Issue #267, opened 2020-06-17
Command Inheritance Extend


Paul-Ferrell:
Make it so that you can extend (and maybe prefix) pavilion commands when using inheritance, rather than just completely overwriting the existing commands.

## Option 1
```yaml

_base:
  run:
    cmds:
      - echo 'hello, punks'

mytest:
  inherits_from: _base
  run:
    cmds_prepend: 
      - echo 'I happen first.'
    cmds_extend:
      - echo 'I happen last.'
```

## Option 2
```yaml
# Using the same base as above.

mytest2:
  inherits_from_base:
  run:
    # Options are append | prepend | replace
    # replace is default.
    cmd_action: append  # The name of this is debatable.
    cmds:
      - echo 'I happen last.'
```

## Option 3
Add '@some_var' syntax to insert a list of variables into place. 

Questions.
 - Complex variables?

Paul-Ferrell:
@kjeverson - We're doing option 1, but I'm changing the names:

`cmds_extend_before`
`cmds_extend_after`

I want the name to call out the the behavior is special here.

Paul-Ferrell:
Merged

PR #294, opened 2020-07-17
Command inheritance extend #267

kjeverson:
Addresses issue #267 - option 1


lib/pavilion/test_config/resolver.py
```diff
@@ -473,6 +476,26 @@ def load_raw_configs(self, tests, host, modes):
 
         return picked_tests
 
+    def cmd_extend_update(self, test_suite_cfg, suite_tests):
```
Paul-Ferrell, line 479: Do this at the level of individual test configs, not at the suite level.

lib/pavilion/test_config/resolver.py
```diff
@@ -655,6 +678,18 @@ def resolve_inheritance(base_config, suite_cfg, suite_path):
...
+                    error = ("Test {} has {} element in the {} section of "
+                             "config, but no inheritance.")
+                    for section in ['build', 'run']:
+                        for cmd in ['cmds_extend_before', 'cmds_extend_after']:
+                            try:
+                                if test_cfg[section][cmd]:
+                                    raise TestConfigError(error.format(
+                                        test_cfg_name, cmd, section))
+                            except KeyError as err:
+                                continue
```
Paul-Ferrell, line 692: This shouldn't be an error.

lib/pavilion/test_config/resolver.py
```diff
@@ -688,6 +723,19 @@ def resolve_inheritance(base_config, suite_cfg, suite_path):
             suite_tests[test_cfg_name] = test_config_loader.merge(parent,
                                                                   test_cfg)
 
+            if parent is not '__base__':
```
Paul-Ferrell, line 726: The section will always be empty for `__base__` anyway.

lib/pavilion/test_config/resolver.py
```diff
@@ -688,6 +723,19 @@ def resolve_inheritance(base_config, suite_cfg, suite_path):
...
+                    if not config:
+                        continue
+                    new_cmd_list = []
+                    if config.get('cmds_extend_before'):
+                        new_cmd_list += config.get('cmds_extend_before')
+                    new_cmd_list += suite_tests[test_cfg_name][section]['cmds']
+                    if config.get('cmds_extend_after'):
+                        new_cmd_list += config.get('cmds_extend_after')
+                    suite_tests[test_cfg_name][section]['cmds'] = new_cmd_list
+
```
Paul-Ferrell, line 738: This is what you should have in a function.

Paul-Ferrell: It applies not just to inheritance, but in all steps of layering. mode -> test -> inheritance -> host -> `__base__`

lib/pavilion/test_config/file_format.py
```diff
@@ -277,6 +277,14 @@ class TestConfigLoader(yc.YamlConfigLoader):
...
                     help_text='The sequence of commands to run to perform '
                               'the build.'),
+                yc.ListElem(
+                    'cmds_extend_before', sub_elem=yc.StrElem(),
+                    help_text='Commands to run before inherited build '
+                              'commands.'),
+                yc.ListElem(
+                    'cmds_extend_after', sub_elem=yc.StrElem(),
+                    help_text='Commands to run after inherited build '
+                              'commands.'),
```
Paul-Ferrell, line 287: It just struck me that the names are terrible.

Let's do `prepend_cmds` and `append_cmds`.

lib/pavilion/test_config/resolver.py
```diff
@@ -1159,3 +1165,21 @@ def resolve_section_values(cls, component, var_man, allow_deferred=False,
...
                                   .format(type(component), component))
+
+def resolve_cmd_inheritance(test_cfg):
+
+    for section in ['build', 'run']:
+        config = test_cfg.get(section)
+        if not config:
+            continue
+        new_cmd_list = []
+        if config.get('cmds_extend_before'):
```
Paul-Ferrell, line 1176: This will still error if the value isn't there, because you'll end up trying to extend a list with `None`.

You want: `config.get('cmds_extend_before', [])`

lib/pavilion/test_config/resolver.py
```diff
@@ -1159,3 +1165,21 @@ def resolve_section_values(cls, component, var_man, allow_deferred=False,
...
+            new_cmd_list += config.get('cmds_extend_before')
+            config['cmds_extend_before'] = []
+        new_cmd_list += test_cfg[section]['cmds']
+        if config.get('cmds_extend_after'):
+            new_cmd_list += config.get('cmds_extend_after')
```
Paul-Ferrell, line 1181: This is fine here, but if you're doing a lot of list extensions (especially with very long lists), you should use the `.extend()` method.
Your way has to copy both lists into a new list, which takes the time and memory equal to the size of both.
`.extend()` just adds to an existing list, and only consumes resources based on the size of what you're adding.

Paul-Ferrell (approved):

Issue #287, opened 2020-07-10
Refactor clean command


Paul-Ferrell:
Part of a general effort to move code from command plugins into the Pavilion core where it's more accessible as part of the Pavilion API. 

 - Add a 'delete' function to the dir_db module that locks the directory and then cleans up test/suite directories based on some filter. 
   - The filter should be a function that takes the path to the directory.
 - Similarly, add code to the 'builder' module to safely delete builds based on a filter function.
    - Make sure builds are only deleted if they aren't being used by a test run and they're at least as old as when you checked if they were in use. (A current bug).

PR #297, opened 2020-07-21
Refactor clean

kjeverson:
addresses issue #287


lib/pavilion/plugins/commands/clean.py
```diff
@@ -46,6 +47,53 @@ def _setup_arguments(self, parser):
...
+        except (commands.CommandError, ValueError) as err:
+            output.fprint("Command Argument Error:",
+                          color=output.RED, file=self.errfile)
+            output.fprint(err)
+            return errno.EINVAL
+
+        if args.verbose:
+            end = '\n'
+        else:
+            end = ''
```
Paul-Ferrell, line 61: Not using a newline will make insanely ugly output. What are you trying to do here?

Paul-Ferrell: And it looks like it only saves you three lines of output. 

Use a carriage return instead: `\r`

Four lines to say that is a bit silly too. 
```python
end = '\n' if args.verbose else '\r'
```

Paul-Ferrell: The output in the delete function can similarly end in `\r` when not verbose, and use the `clear` option to fprint.

lib/pavilion/plugins/commands/clean.py
```diff
@@ -46,6 +47,53 @@ def _setup_arguments(self, parser):
...
+        # Clean Tests
+        removed_tests = 0
+        tests_dir = pav_cfg.working_dir / 'test_runs'     # type: Path
+        output.fprint("Removing Tests...", file=self.outfile, end=end)
+        removed_tests = dir_db.delete(pav_cfg, tests_dir, args)
+        output.fprint("Removed {} test(s).".format(removed_tests),
+                      file=self.outfile, color=output.GREEN)
+
+        # Clean Series
+        removed_series = 0
```
Paul-Ferrell, line 72: You overwrite this just a few lines down.

lib/pavilion/plugins/commands/clean.py
```diff
@@ -46,6 +47,53 @@ def _setup_arguments(self, parser):
...
+            output.fprint(err)
+            return errno.EINVAL
+
+        if args.verbose:
+            end = '\n'
+        else:
+            end = ''
+
+        # Clean Tests
+        removed_tests = 0
```
Paul-Ferrell, line 64: You overwrite this just a few lines down

lib/pavilion/plugins/commands/clean.py
```diff
@@ -46,6 +47,53 @@ def _setup_arguments(self, parser):
...
+        # Clean Series
+        removed_series = 0
+        series_dir = pav_cfg.working_dir / 'series'       # type: Path
+        output.fprint("Removing Series...", file=self.outfile, end=end)
+        removed_series = dir_db.delete(pav_cfg, series_dir, args)
+        output.fprint("Removed {} series.".format(removed_series),
+                      file=self.outfile, color=output.GREEN)
+
+        # Clean Builds
+        removed_builds = 0
```
Paul-Ferrell, line 80: here too

lib/pavilion/plugins/commands/clean.py
```diff
@@ -73,146 +121,20 @@ def run(self, pav_cfg, args):
...
                 date = ' '.join(args.older_than)
                 try:
                     cutoff_date = datetime.strptime(date, '%b %d %Y')
-                except (TypeError, ValueError):
-                    output.fprint("{} is not a valid date."
-                                  .format(args.older_than),
-                                  file=self.errfile, color=output.RED)
-                    return errno.EINVAL
+                except ValueError as err:
+                    raise ValueError("{} is not a valid date: {}".format(date,
```
Paul-Ferrell, line 130: Start formats on the next line, especially if their arguments are going to wrap.

Paul-Ferrell: Wait, what are you doing? 
Why are you raising exceptions instead of returning a reasonable error code? 

It was correct before.

lib/pavilion/plugins/commands/clean.py
```diff
@@ -73,146 +121,20 @@ def run(self, pav_cfg, args):
...
+                                                                        err))
             else:
-                output.fprint(
-                    "Invalid `--older-than` value.", file=self.errfile,
-                    color=output.RED
```
Paul-Ferrell, line 88: Same here.

Paul-Ferrell: Now, if you move functionality into other functions than `run`, you have to pass up error information somehow. Doing so with an exception is generally reasonable, but you should have an exception specific to the module to use and catch.

lib/pavilion/clean.py
```diff
@@ -0,0 +1,65 @@
...
+
+        try:
+            test_time = datetime.fromtimestamp(_.lstat().st_mtime)
+        except FileNotFoundError:
+            return False
+
+        if test_time > cutoff_date:
+            return False
+
+        complete_path = _/'RUN_COMPLETE'
```
Paul-Ferrell, line 25: Use a full variable name. This isn't perl.

lib/pavilion/clean.py
```diff
@@ -0,0 +1,65 @@
...
+        still exist.
+        :param _: This is a passed path object.
+        :return True: If series dir can be removed.
+        :return False: If series dir cannot be removed.
+        """
+        path = _
+        for test_path in path.iterdir():
+            if (test_path.is_symlink() and
+                test_path.exists() and
+                test_path.resolve().exists()):
```
Paul-Ferrell, line 61: Again, be careful with this. Look in utils.py for a safer variant.

lib/pavilion/dir_db.py
```diff
@@ -92,24 +97,73 @@ def create_id_dir(id_dir: Path) -> (int, Path):
...
+        int(string, 16)
+    except ValueError as err:
+        return False
+
+    return True
 
 def select(id_dir: Path,
-           filter_func: Callable[[Path], bool] = default_filter) -> List[Path]:
+           filter_func: Callable[[Path], bool] = default_filter,
+           builds_dir: bool = False) -> List[Path]:
```
Paul-Ferrell, line 120: Don't tie this to the idea that it's a build directory, make it a generic flag to indicate hex filenames.

lib/pavilion/dir_db.py
```diff
@@ -92,24 +97,73 @@ def create_id_dir(id_dir: Path) -> (int, Path):
...
-            if filter_func(path):
-                passed.append(path)
+        if not builds_dir:
+            if path.name.isdigit() and path.is_dir():
+                if filter_func(path):
+                    passed.append(path)
+        else:
+            if is_hex(path.name) and path.is_dir():
+                if filter_func(path):
+                    passed.append(path)
```
Paul-Ferrell, line 135:
```python
fn_base = 16 if hex_fn else 10
for path in id_dir.iterdir():
  try:
    int(path.name, fn_base)
  except ValueError:
    continue
  
  # Then you only have to check that it's a directory and the filter function
```

Paul-Ferrell: In fact, you could just make fn_base the function argument instead of build_dir (and default it to 10).

lib/pavilion/dir_db.py
```diff
@@ -92,24 +97,73 @@ def create_id_dir(id_dir: Path) -> (int, Path):
...
+        for path in select(id_dir, filter_func):
+            try:
+                shutil.rmtree(path)
+            except OSError as err:
+                output.fprint("Could not remove {} {}: "
+                              "{}".format(id_dir.name, path, err),
+                              color=output.YELLOW)
+                continue
+            count += 1
+            if verbose:
```
Paul-Ferrell, line 163: Again, return a list of messages along with the count, and let the caller deal with them. 

I acknowledge that this is awkward, but it makes usage outside of commands (like with unit-testing) a lot easier.

lib/pavilion/builder.py
```diff
@@ -1014,3 +1011,84 @@ def __eq__(self, other):
...
+
+        :return True: The passed build path can be removed.
+        :return False: The passed build path cannot be removed.
+
+        """
+        return not path.name in used_build_paths
+
+    count = 0
+
+    lock_path = builds_dir.with_suffix('.lock;')
```
Paul-Ferrell, line 1059: Why the semi-colon?

lib/pavilion/builder.py
```diff
@@ -1014,3 +1011,84 @@ def __eq__(self, other):
...
+                error_list.append("Could not remove build {}: {}"
+                                  .format(path, err))
+                continue
+            count += 1
+            if verbose:
+                output.fprint('Removed build {}.'.format(path.name))
+
+    return count, error_list
+
+def remove_lingering_finished_files(builds_dir: Path) -> None:
```
Paul-Ferrell, line 1076: I think we can do without this. You remove them above when you remove the build. I just need to change how new builds handle a missing build when .finished exists.

lib/pavilion/dir_db.py
```diff
@@ -92,24 +97,72 @@ def create_id_dir(id_dir: Path) -> (int, Path):
...
+                error_list.append("Could not remove {} {}: {}"
+                                  .format(id_dir.name, path, err))
+                continue
+            count += 1
+            if verbose:
+                output.fprint("Removed {} {}.".format(id_dir.name,
+                                                      path.name))
+
+    reset_pkey(id_dir)
+    return count, error_list
```
Paul-Ferrell, line 167: Return a list of removed directory paths, and a list of errors. Get rid of all prints.

Paul-Ferrell:
We're going to have to revisit this eventually given what Calvin and I have done with filtering. It won't change much though.

Oh, one more thing. I think you should move the build deletion stuff into the clean.py library you added.

Paul-Ferrell (approved):

Paul-Ferrell (approved):

Issue #296, opened 2020-07-21
add "--save" option to the "pav results -r testid" command


tygoetsch:
Currently we can tell pavilion to reparse a test using the most recent result_parse section of a test, however the results from that command, `pav results -r testid` are not saved to disk. It would be useful to add an option to carry out this function: `pav results -r --save testid` or something similar.

kjeverson:
PR #298 addressed this and has been merged.

PR #298, opened 2020-07-22
Initial re-run and save option for result command.

kjeverson:
addresses issues #296

### DIFF
--- a/lib/pavilion/plugins/commands/result.py
+++ b/lib/pavilion/plugins/commands/result.py
@@ -65,7 +65,8 @@ class ResultsCommand(commands.Command):
         parser.add_argument(
             '-s', '--save',
             action='store_true', default=False,
-            help="Save the the re-run to the Test's results file."
+            help="Save the re-run to the test's results json and log. Will "
+                 "not update the general pavilion result log."
         )
         parser.add_argument(
             '-l', '--show-log', action='store_true', default=False,
@@ -100,9 +101,6 @@ class ResultsCommand(commands.Command):
         if args.re_run:
             if not self.update_results(pav_cfg, tests, log_file):
                 return errno.EINVAL
-
-        if args.show_log and args.save:
-            log_file = io.StringIO()
 
         if args.save:
             if not self.update_results(pav_cfg, tests, log_file, save=True):
@@ -211,8 +209,8 @@ class ResultsCommand(commands.Command):
         :param pav_cfg: The pavilion config.
         :param tests: A list of test objects to update.
         :param log_file: The logfile to log results to. May be None.
-        :param save: A bool that determines if the updated results will be
-                     saved or not. By default it is always false.
+        :param save: Whether to save the updated results to the test's result
+                     log. It will not update the general result log.
         :returns: True if successful, False otherwise. Will handle
             printing of any failure related errors.
         """
--- a/lib/pavilion/plugins/commands/result.py
+++ b/lib/pavilion/plugins/commands/result.py
@@ -53,8 +53,7 @@ class ResultsCommand(commands.Command):
             "-f", "--full", action="store_true", default=False,
             help="Show all result keys."
         )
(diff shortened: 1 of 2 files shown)


lib/pavilion/plugins/commands/result.py
```diff
@@ -53,15 +53,22 @@ def _setup_arguments(self, parser):
...
             action='store_true', default=False,
-            help="Re-run the results based on the latest version of the test"
+            help="Re-run the results based on the latest version of the test "
                  "configs, though only changes to the 'result' section are "
                  "applied. This will not alter anything in the test's run "
                  "directory; the new results will be displayed but not "
                  "otherwise saved or logged."
         )
+        rerun_group.add_argument(
+            '-rs', '--re-run-save', dest="re_save",
```
Paul-Ferrell, line 67: Just name this '-s', '--save'

lib/pavilion/plugins/commands/result.py
```diff
@@ -191,7 +205,7 @@ def _get_tests(self, pav_cfg, tests_arg):
         return map(int, test_list)
 
     def update_results(self, pav_cfg: dict, tests: List[TestRun],
-                       log_file: IO[str]) -> bool:
+                       log_file: IO[str], save=False) -> bool:
```
Paul-Ferrell, line 208: Add the type hint for save, `save: bool = False`, and add it to the params in the doc string.

lib/pavilion/plugins/commands/result.py
```diff
@@ -278,7 +292,16 @@ def update_results(self, pav_cfg: dict, tests: List[TestRun],
...
             # The new results will be attached to the test (but not saved).
-            test.gather_results(test.results.get('return_value', 1),
-                                regather=True, log_file=log_file)
+            results = test.gather_results(test.results.get('return_value', 1),
+                                          regather=True, log_file=log_file)
+
+            if save:
+                results_log = test.path/'results.log'
+                test.save_results(results)
+                with results_log.open('a') as log:
```
Paul-Ferrell, line 301: You don't need to reopen the log file and append to it; because if you're saving `log_file` should already be that file, open.  Also, I'm pretty sure there's a 'result_log_path' variable on the test object already.

kjeverson: It looks like the log_file that is passed to this method is used for the command argument `--show-log`, and is either None or a StringIO. It isn't a test object's log file. Furthermore, the update_results method takes a list of Test_Run objects and iterates over them, because of this I have to open the results_log for each test instance...

lib/pavilion/plugins/commands/result.py
```diff
@@ -96,6 +101,13 @@ def run(self, pav_cfg, args):
             if not self.update_results(pav_cfg, tests, log_file):
                 return errno.EINVAL
 
+        if args.show_log and args.save:
+            log_file = io.StringIO()
```
Paul-Ferrell, line 105: You shouldn't require saving to show the log.

lib/pavilion/plugins/commands/result.py
```diff
@@ -191,14 +203,16 @@ def _get_tests(self, pav_cfg, tests_arg):
...
+                       log_file: IO[str], save: bool = False) -> bool:
         """Update each of the given tests with the result section from the
         current version of their configs. Then rerun result processing and
         update the results in the test object (but change nothing on disk).
 
         :param pav_cfg: The pavilion config.
         :param tests: A list of test objects to update.
         :param log_file: The logfile to log results to. May be None.
+        :param save: A bool that determines if the updated results will be
+                     saved or not. By default it is always false.
```
Paul-Ferrell, line 215: You don't need to restate the type.
'Whether to save the updated results to the test's result log.' 
You should check if this also saves them to the general result log, and comment about that. I don't care if it does or not, but the argument here and in the command arguments should say.

Issue #309, opened 2020-08-20
Add option to (partly) flatten results when they go into the general result log.


Paul-Ferrell:
1. Unify 'fn' and 'n' result fields into a single 'by_file' dict.
2. Add a pavilion.yaml option (default True), 'flatten_results' results.

When `flatten_results` is True, transform the final results for each test into multiple result records, one for each 'by_file' value.
1. There should be a 'file' reserved key for the 'by_file' key.
2. The values in the by_file key should be merged with the base result dict.
3. Each such record should be added as a seperate log line in the results log.

PR #310, opened 2020-08-20
Add result flattening for the general result log. On by default.

Paul-Ferrell:


Paul-Ferrell:
Closes #309

lib/pavilion/test_run.py
```diff
@@ -781,15 +780,31 @@ def gather_results(self, run_result, regather=False, log_file=None):
...
+            base = results.copy()
+            del base['per_file']
+
+            for per_file, values in results['per_file'].items():
+                per_result = base.copy()
+                per_result['file'] = per_file
+                per_result.update(values)
+
+                result_logger.info(output.json_dumps(per_result))
+        else:
```
nicholas-sly, line 805: Is this strictly an either-or situation, or should this be done in both cases (minus the per_file content)?

test/tests/result_tests.py
```diff
@@ -544,4 +544,43 @@ def test_re_search(self):
         for key, answer in answers.items():
             self.assertEqual(results[key], answer)
 
+    def test_flatten_results(self):
```
nicholas-sly, line 547: If this functionality should support updating per_file results as well as a global result, that should be tested here as well.

lib/pavilion/test_run.py
Paul-Ferrell: If there's no per_file content, there's nothing to flatten.

test/tests/result_tests.py
Paul-Ferrell: I extended the test.

Issue #292, opened 2020-07-17
Add a run/build option to specify a different file to follow for run/build timeouts.


Paul-Ferrell:
```yaml

mytest:
  run:
    # mytest sends output to mytest.out
    cmds: '{{test_cmd}} ./mytest'
    timeout_file: mytest.out
```

Unlike with the run log, 'mytest.out' need not exist the entire time.

kjeverson:
How should I handle if we get to the timeout check and the file doesn't exist? Also, what should I do if the file just disappears? I can put it in a try-except block and catch the `FileNotFoundError`, but should I then look back at the default build/run.log or should there be an error raised...

PR #316, opened 2020-08-27
Timeout file

kjeverson:
Addresses issue #292

### DIFF
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -592,15 +592,26 @@ class TestBuilder:
                                         stderr=build_log)
 
                 result = None
+                start = time.time()
+                deleted = None
+                file_created = False
                 while result is None:
                     try:
                         result = proc.wait(timeout=1)
                     except subprocess.TimeoutExpired:
                         try:
                             log_stat = self._timeout_file.stat()
+                            file_created = True
+                            timeout = log_stat.st_mtime + self._timeout
                         except FileNotFoundError as err:
-                            continue
-                        timeout = log_stat.st_mtime + self._timeout
+                            # Timeout file has not been created yet.
+                            if not file_created:
+                                timeout = start + self._timeout
+                            # Timeout file existed, but was deleted.
+                            else:
+                                if deleted is None:
+                                    deleted = time.time()
+                                timeout = deleted + self._timeout
                         # Has the output file changed recently?
                         if time.time() > timeout:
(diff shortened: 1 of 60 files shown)


lib/pavilion/builder.py
```diff
@@ -586,12 +592,26 @@ def _build(self, build_dir, cancel_event):
...
+                            timeout = log_stat.st_mtime + self._timeout
+                        except FileNotFoundError as err:
+                            # Timeout file has not been created yet.
+                            if not file_created:
+                                timeout = start + self._timeout
+                            # Timeout file existed, but was deleted.
+                            else:
+                                if deleted is None:
+                                    deleted = time.time()
+                                timeout = deleted + self._timeout
```
Paul-Ferrell, line 614: Let's simplify the problem a bit. 

1. If the timeout file doesn't exist, use the log file.
2. `timeout = max(timeout, stat.st_mtime + self._timeout)`

Then you don't have to track creation or deletion, and your only conditional is `if self._timeout_file.exists(): timeout_file = ...`

lib/pavilion/builder.py
```diff
@@ -586,12 +592,20 @@ def _build(self, build_dir, cancel_event):
...
-                        log_stat = self.tmp_log_path.stat()
-                        timeout = log_stat.st_mtime + self._timeout
+                        if self._timeout_file.exists():
+                            log_stat = self._timeout_file.stat()
+                            timeout = max(timeout, log_stat.st_mtime +
+                                          self._timeout)
+                        else:
+                            log_stat = self.tmp_log_path.stat()
+                            timeout = max(timeout, log_stat.st_mtime +
+                                          self._timeout)
```
Paul-Ferrell: There's a race condition here that I just noticed. The file could disappear between checking for its existence and checking 'stat'. It's not likely, but if it happens things will explode.

```python
timeout_file = self.timeout_file if self.timeout_file.exists else self.tmp_log_path
try:
    timeout = max(timeout_file.stat().st_mtime + self._timeout,
                            timeout)
except OSError:
    pass
```

lib/pavilion/builder.py
Paul-Ferrell: I tweaked your fix to this a little bit, and addressed a race condition in the process.

Paul-Ferrell (approved):
Ready as soon as the tests pass.

PR #319, opened 2020-09-09
Result filtering

Paul-Ferrell:
Addresses several issues:

  - The 'result' command now has full filtering capability like 'status' and 'list'.
  - Added the 'maint' command as a place to put commands to tweak pav operations.
  - Added the 'maint prune_results' command to allow for result deletion from the main result log.
  - Swapped 'sched.test_node_list' for 'sched.test_node_list_short' in the results of slurm based tests.


kjeverson (approved):
Checked it out, everything passed on my end and the commands you changed/added seem to work as expected. Look good to me.

PR #333, opened 2020-10-20
2.2 bugfixes 3

Paul-Ferrell:


kjeverson (approved):
Nothing stood out to me. Everything seemed to function as it has previously (nothing broke) and all unit tests passed.

PR #335, opened 2020-10-23
Finished the scheduler plugin documentation.

Paul-Ferrell:


docs/plugins/schedulers.rst
```diff
@@ -1,46 +1,791 @@
...
+
+    # Building your commands with a list is simple and flexible.
+    if redirect_output:
+        my_cmd.extend(['-o', outfile])
+
+    mycmd.append('--partition=strange')
+
+    subprocess.run(my_cmd)
+
+To find commands on a system, the
```
nicholas-sly, line 419: You've wandered off again.  I think this is covered at the top of this subsection.

docs/plugins/schedulers.rst
```diff
@@ -1,46 +1,791 @@
...
+You can also add to the environment through the ``env`` argument, though you
+need to make sure to include the base environment in most cases.
+
+.. code-block:: python
+
+    import os
+    import subprocess
+
+    myenv = dict(os.environ)
+    myenv['MY_ENV_VAR'] = 'Hiya!'
```
nicholas-sly, line 433: Might add an example of adding something arbitrary to PATH to drive home your point about including the base environment.

docs/plugins/schedulers.rst
```diff
@@ -1,46 +1,791 @@
...
+
+Additionally, there are four basic cases that need to be handled:
+
+1. The job was never started. This is handled for you in ``cancel_job()``,
+   which calls ``_cancel_job()``.
+2. The job is enqueued but not yet running (or somewhere in between).
+3. The job is running.
+4. The job has finished.
+
+Most of the time, this simply means you will try to cancel the job id,
```
nicholas-sly, line 517: No comma at the end.

docs/plugins/schedulers.rst
```diff
@@ -1,46 +1,791 @@
...
+and capture any errors.
+
+Additionally, if your job id encodes information that could denote that
```
nicholas-sly, line 520: Might be a good place to reference a concept that we've allegedly been looking into of having a central place to submit jobs to any machine.  Could involve having more information for your `scancel` call.

```
-bash-4.2$ scancel --help
Usage: scancel [OPTIONS] [job_id[_array_id][.step_id]]
  -A, --account=account           act only on jobs charging this account
  -b, --batch                     signal batch shell for specified job
  -f, --full                      signal batch shell and all steps for specified job
  -H, --hurry                     avoid burst buffer stage out
  -i, --interactive               require response from user for each job
  -M, --clusters                  clusters to issue commands to.
                                  NOTE: SlurmDBD must be up.
  -n, --name=job_name             act only on jobs with this name
  -p, --partition=partition       act only on jobs in this partition
  -Q, --quiet                     disable warnings
  -q, --qos=qos                   act only on jobs with this quality of service
  -R, --reservation=reservation   act only on jobs with this reservation
      --sibling=cluster_name      remove an active sibling job from a federated job
  -s, --signal=name | integer     signal to send to job, default is SIGKILL
  -t, --state=states              act only on jobs in this state.  Valid job
                                  states are PENDING, RUNNING and SUSPENDED
  -u, --user=user_name            act only on jobs of this user
  -V, --version                   output version information and exit
  -v, --verbose                   verbosity level
  -w, --nodelist                  act only on jobs on these nodes
      --wckey=wckey               act only on jobs with this workload
                                  charactization key

Help options:
  --help                          show this help message
  --usage                         display brief usage message
```

nicholas-sly (changes requested):
Few little nitpicks, then it'll be good to go.

PR #336, opened 2020-10-26
Added support for configuring the job name in a test.

hugegreenbug:
I have a need for specifying a job name in a test. This patch adds that as a config file option.


Paul-Ferrell (approved):

lib/pavilion/plugins/sched/slurm.py
```diff
@@ -45,9 +45,15 @@ def get_lines(self):
...
-            .format(s=self))
+         if self._conf.get('job_name') is not None:
+             lines.append(
+                 '#SBATCH --job-name {s._conf[job_name]}'
+                 .format(s=self))
+         else:
+             lines.append(
+                 '#SBATCH --job-name "pav test #{s._test_id}"'
+                 .format(s=self))
+
```
Paul-Ferrell, line 56: The indentation is off here, which is breaking the whole plugin. Other than that this looks fine.

Paul-Ferrell (changes requested):

Paul-Ferrell (approved):
Looks good. Merging.

lib/pavilion/plugins/sched/slurm.py
hugegreenbug: fixed

PR #342, opened 2020-11-05
added `pav wait --summary`, which gives a count for every different state in one line 

### LINKED ISSUE #291: Summary option for pav wait
Add an option to the `pav wait` command to print a summary of the status of the latest test series on a single line and update it, much the same as the build section does.

francinelapid:
Example: `COMPLETE: 1 | RUNNING: 1`

It won't say whether a test passed or failed, but I can add that if we want. 

Closes #291


lib/pavilion/plugins/commands/wait.py
```diff
@@ -80,37 +103,47 @@ def run(self, pav_cfg, args):
...
+                stats_out = []
+
+                if out_mode == self.OUT_SILENT:
+                    pass
+                elif out_mode == self.OUT_ONELINE:
+                    stats_out.extend([str(time.ctime(time.time())), ': '])
+                    for test in stats:
+                        stat = "{t[test_id]}({t[name]}) {t[state]} | "\
+                            .format(t=test)
+                        stats_out.append(stat)
```
Paul-Ferrell, line 119: - Use the time shortening function in output/utils that Jordan wrote. 
- Instead of giving a single message per test, organize by the current test status like the one-line status does when building.  `RUNNING: 5 | COMPLETE: 5 | ` etc...
- Truncate that at the width of the terminal. Terminal width is available through `shutil`.

Paul-Ferrell (commented):
I made some code simplifications and cleaned up a few things. I (mostly) left your additions alone though. See above/below for the formatting changes I requested.

lib/pavilion/plugins/commands/wait.py
Paul-Ferrell: Apparently that's what '--summary' already does.  Nevermind.

Paul-Ferrell (approved):
Turns out the code I was commenting on wasn't yours. You did what I expected in the first place. 
I removed the confusingly similar '--oneline' and kept your version.

PR #345, opened 2020-11-17
Permissions issues2

Paul-Ferrell:


Paul-Ferrell:
- Executable permissions for extracted zips fixed
 - General permissions bugs addressed
 - The general pav.log is now a single file for simplicity.
 - Greatly expanded permission unit tests (mostly by making them very strict).
 - Refactored the extraneous prints test.

lib/pavilion/scriptcomposer.py
```diff
@@ -9,6 +9,8 @@
 from pavilion import module_wrapper
 from pavilion.module_actions import ModuleAction
 
+from pavilion.permissions import PermissionsManager
```
nicholas-sly, line 12: This doesn't appear to be used anywhere in this file...

lib/pavilion/test_run.py
```diff
@@ -574,14 +576,18 @@ def get_permissions(pav_cfg, config) -> (str, int):
...
-        umask = config.get('umask', pav_cfg['umask'])
+        umask = config.get('umask')
+        if umask is None:
+            umask = pav_cfg['umask']
         if umask is not None:
             try:
                 umask = int(umask, 8)
             except ValueError:
                 raise RuntimeError(
                     "Invalid umask. This should have been enforced by the "
```
nicholas-sly, line 587: Not part of your changes, but this is a little repetitive.

nicholas-sly (changes requested):
Noticed a couple of things.  Otherwise, looks good.

lib/pavilion/test_run.py
Paul-Ferrell: It's not. config['umask'] can have a value of `None` in that dictionary. It was a bug I ran across.

PR #347, opened 2020-11-20
Lockfile Streamlining

Paul-Ferrell:
The cross-system lockfile tests ran fine.

aarch systems are really slow at getting update nfs metadata, as usual, so they're only able to acquire a lock after any contention has been resolved.


francinelapid (approved):
Looked through it and ran a simple test.

PR #348, opened 2020-11-25
Add .circleci/config.yml

Paul-Ferrell:


francinelapid (approved):
table width corrections look fine

PR #352, opened 2020-12-04
Clean removes lingering build files.

### LINKED ISSUE #350: Build's .log and .lock files don't get removed with Pav Clean
Killed build's leave lingering `.log` and `.lock` files in the builds directory. Pav clean never attempts to remove them. Can cause `BUILD_WAIT` and locking issues when the same test is ran later on.

kjeverson:
closes #350 

Removes any files in the build dir that are not and cannot be handled when removing build dirs (essentially `.lock` and `.log` files from killed builds.)


lib/pavilion/clean.py
```diff
@@ -104,3 +105,34 @@ def _get_used_build_paths(tests_dir: Path) -> set:
...
+        if path.is_dir():
+            continue
+        # Don't remove anything associated with a used build hash
+        if path.stem in used_build_paths:
+            continue
+
+        path.unlink()
+        if verbose:
+            msgs.append("Removed lingering build file {}."
+                        .format(path.name))
```
Paul-Ferrell, line 137: This would delete anything else that we ever happen to put in this directory. If you're concerned about locks and logs, then specifically delete those.

Paul-Ferrell (changes requested):

Paul-Ferrell (approved):

PR #353, opened 2020-12-08
Fixed rename bug.

Paul-Ferrell:
I introduced a bug that's only in slurm...


kjeverson (approved):
Somewhere along the way we lost the  "1 Tests started as series S1" message.
```
-bash-4.2$ pav run branson
Resolving Test Configs: 100%
Creating Test Runs: 100%
BUILD_REUSED: 1   
-bash-4.2$      
```

PR #356, opened 2020-12-09
pav log series s<id>, hide _series from pav --help

francinelapid:
1. added `pav log series s<id>` 
2. hide `_series` from `pav --help`
3. added a second to the sleep function call in the `series_simultaneous` test 
4. BUG: fixed modes handling issue

### DIFF
--- a/test/tests/series_tests.py
+++ b/test/tests/series_tests.py
@@ -60,8 +60,6 @@ class SeriesFileTests(PavTestCase):
 
         test_starts = []
         for test_id, test_obj in test_series_obj.tests.items():
-            from pavilion import output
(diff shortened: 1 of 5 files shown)


test/tests/series_tests.py
```diff
@@ -56,7 +56,7 @@ def test_series_simultaneous(self):
         test_series_obj.run_series()
 
         # make sure test actually ends
-        time.sleep(2)
+        time.sleep(3)
```
Paul-Ferrell, line 59: I think we need to add a 'wait' method, but this is fine for now.

Paul-Ferrell (approved):

PR #360, opened 2020-12-16
fixed series outfile bug

francinelapid:
This shows the `4 tests started as test series s2.` line when I'm starting tests via `pav run`.


kjeverson (approved):
Looks like this works

PR #371, opened 2021-01-15
Status history

### LINKED ISSUE #370: Add pav status --history.
- Should require a single test be listed. 
 - Should print every status change for that single test. 
 - Should be implemented mostly in status_utils

kjeverson:
closes #370

### DIFF
--- a/lib/pavilion/plugins/commands/status.py
+++ b/lib/pavilion/plugins/commands/status.py
@@ -50,10 +50,9 @@ class StatusCommand(commands.Command):
         series."""
 
         try:
-            test_ids = cmd_utils.arg_filtered_tests(pav_cfg, args)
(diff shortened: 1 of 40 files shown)


lib/pavilion/plugins/commands/status.py
```diff
@@ -37,70 +37,41 @@ def _setup_arguments(self, parser):
             '-s', '--summary', default=False, action='store_true',
             help='Display a single line summary of test statuses.'
         )
+        parser.add_argument(
+            '--history', default=False, action='store_true',
+            help='Display status history for a single test_run.'
+        )
```
Paul-Ferrell, line 43: Create a 'mutually exclusive argument group for --summary and --history, since they are mutually exclusive.

Paul-Ferrell (changes requested):
Huh, it looks like we already had a function to display the history, but it just wasn't used...

Paul-Ferrell (approved):

PR #372, opened 2021-01-15
Pav run --hard-fail

### LINKED ISSUE #367: Add '--hard-fail' to `pav run`.
Right now Pavilion cancels all started tests when any one of them fail to start or build. 

Change this behavior so that it prints out an error, but continues with the rest by default. 
 - This may require changes to the scheduler plugins as well, as I think they take a list of tests to start. 
 - The '--hard-fail' option would thus revert to the old behavior.

kjeverson:
closes #367

### DIFF
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -373,6 +373,11 @@ class TestBuilder:
                               .format(self.name), state=STATES.ABORTED)
             return False
 
+        if cancel_event.is_set():
+            self.tracker.fail("Build aborted due to failures in other builds.",
+                              states=STATES.ABORTED)
+            return False
+
         # Only try to do the build if it doesn't already exist and is finished.
         if not self.finished_path.exists():
             # Make sure another test doesn't try to do the build at
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -362,7 +362,7 @@ class TestBuilder:
         """
 
         if fail_event.is_set():
-            self.tracker.fail("Run aborted due to failures in build '{}'."
+            self.tracker.fail("Build aborted due to failures in build '{}'."
                               .format(self.name), state=STATES.ABORTED)
             return False
 
@@ -445,7 +445,7 @@ class TestBuilder:
 
                 else:
                     if fail_event is not None and fail_event.is_set():
-                        self.tracker.fail("Run aborted due to failures in build '{}'."
+                        self.tracker.fail("Build aborted due to failures in build '{}'."
                                           .format(self.name), state=STATES.ABORTED)
                         return False
                     else:
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -52,7 +52,6 @@ class TestBuilder:
 
     DEPRECATED = ".pav_deprecated_build"
     FINISHED_SUFFIX = '.finished'
-    FAILED_SUFFIX = '.failed'
 
     LOG_NAME = "pav_build_log"
 
@@ -102,7 +101,6 @@ class TestBuilder:
         fail_name = 'fail.{}.{}'.format(self.name, self.test.id)
         self.fail_path = pav_cfg.working_dir/'builds'/fail_name
         self.finished_path = self.path.with_suffix(self.FINISHED_SUFFIX)
-        self.failed_path = self.path.with_suffix(self.FAILED_SUFFIX)
 
         if self._timeout_file is not None:
             self._timeout_file = self.path/self._timeout_file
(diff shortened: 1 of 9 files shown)


lib/pavilion/cmd_utils.py
```diff
@@ -380,7 +381,7 @@ def build_local(tests: List[TestRun],
 
         test_threads = [thr for thr in test_threads if thr is not None]
 
-        if cancel_event.is_set():
+        if cancel_event.is_set() and hard_fail is True:
```
Paul-Ferrell, line 384: So... this won't work. 

Once the cancel event is set, all the builds with break regardless of if hard_fail is true. Builds watch for the cancel event and quit if they see it set. They also set it on failure.  

I actually came to annotate the ticket with info on how to deal with this. You need to:
1. Know when a build failed.
2. Have the primary thread cancel the builds when one fails, but only if hard_fail is True.

Currently both these steps are handled by the cancel event without intervention by the main thread. This will have to be separated. 
You can:
1. Have two events. A cancel_event to tell the builds to stop and a fail_event for the builds to tell the main thread they failed. 
2. Have one event still, but tell the main thread through other means such as by modifying the build info tracker.
    - You could also track this with the status file, but I don't recommend it.
3. Something more clever.

Paul-Ferrell (changes requested):

lib/pavilion/builder.py
```diff
@@ -567,8 +575,10 @@ def _build(self, build_dir, cancel_event, lock: lockfile.LockFile = None):
                             return False
 
         except subprocess.CalledProcessError as err:
-            if cancel_event is not None:
+            if cancel_event is not None and hard_fail:
                 cancel_event.set()
```
Paul-Ferrell, line 579: Only the master thread needs to know about hard fail and set the cancel event. It should be watching for the the fail event, and then use that to decide whether to cancel.

Paul-Ferrell (changes requested):

Paul-Ferrell:
Make sure to cancel all builds that would be waiting on the one that failed.

lib/pavilion/cmd_utils.py
```diff
@@ -380,35 +386,23 @@ def build_local(tests: List[TestRun],
...
-                    "the test build log (pav log build {id})"
-                    .format(id=failed_build.test.id), file=errfile)
-
-            return errno.EINVAL
+                for test in tests:
+                    if (test.status.current().state not in
+                            (STATES.BUILD_FAILED, STATES.BUILD_ERROR, STATES.BUILD_TIMEOUT)):
+                        test.status.set(
+                            STATES.ABORTED,
+                            "Run aborted due to failures in other builds.")
```
Paul-Ferrell, line 403: Cancel based on build hash, not status file.

lib/pavilion/builder.py
```diff
@@ -352,15 +352,24 @@ def _update_src(self):
...
-        :param threading.Event cancel_event: Allows builds to tell each other
-        to die.
+        :param threading.Event cancel_event: Signals build to stop.
+        :param threading.Event fail_event: Allows build to report failure,
+        without cancelling other builds.
         :return: True if these steps completed successfully.
         """
 
+        if fail_event is None:
+            fail_event = threading.Event()
```
Paul-Ferrell, line 366: Do this with the cancel event to. 

Drop checks for None and if it was already set.

Paul-Ferrell (changes requested):

lib/pavilion/cmd_utils.py
kjeverson: This doesn't actually cancel tests, it sets the proper `ABORTED` status for test/threads that don't get started prior to the cancel event. That being said, the only status I need to actually check for here is `STATE.CREATED` and I did that.

kjeverson:
I think this is ready for review again, I made a new ticket for all the cancel/slurm changes as that might be a substantial overhaul not really relevant to this branch.

Paul-Ferrell (approved):

Paul-Ferrell (changes requested):
Unapproving. 

You've added 'fail.set()' a bunch of times to _build(). Is there a case where _build returns False but did not fail? If not, then you can simply set 'fail' once depending on _build's return values.

Also, you check fail inside of _build. I can see checking it in build; you want to make sure that a different instance of the same build didn't fail before you proceed. Once you've acquired the build lock, you can assume that fail won't change. Checking the cancel event in the middle of _build does make sense though, as that can change in response to another non-conflicting build.

kjeverson:
I think I will actually do all the cancel changes in here, they aren't as substantial as I thought previously.

kjeverson:
Got it all copied over to a new branch and managed to reintroduce the `BUILD_CREATED` bug...

kjeverson:
moved to PR #402

PR #385, opened 2021-02-18
Fixed issue with open parens and whitespace

Paul-Ferrell:
Whitespace was not being ignored before parenthesis. This was originally intentional; open parens
for functions didn't want whitespace between the function name and paren. Unfortunately that
also caused problems for statements like 'True and (False or True)'.

Now whitespace is ignored before parens in general.


francinelapid (approved):
Code looks good. Tested it and it works.

PR #388, opened 2021-02-19
Added additional slurm arguments for mpi and srun_extra

Paul-Ferrell:


Paul-Ferrell:
Added Francine as a reviewer because I know she'll actually verify that this works.

francinelapid (approved):
Code looks good. Tested new config elements and they work.

Issue #165, opened 2020-03-09
pav run count (repeat) option


Paul-Ferrell:
Add an option to have pav run create 'N' instances of the given tests. 

```
pav run --repeat 5 supermagic imb.imb
```
Would run both the supermagic and imb tests 5 times.

CalvinDSeamons:
Kody hungers for tickets... he stole this from me while I was in my stricken state of delirium last Thursday.  Expect a PR from him, I'm just going to add him as a assignee as well.

PR #389, opened 2021-02-19
Run repeat

kjeverson:
closes issue #165 

Introduces a `--repeat` flag, but also supports a `*` notation to repeat tests.

i.e. `pav run test*3` will run `test` 3 times,  as will `pav run --repeat 3 test`


lib/pavilion/plugins/commands/run.py
```diff
@@ -196,11 +204,21 @@ def _get_tests(self, pav_cfg, args, mb_tracker, build_only=False,
...
+        repeat_re = re.compile(r'(.*)\*(\d+)')
+
+        tests = []
+        for test_name in args.tests:
+            match = re.search(repeat_re, test_name)
+            if match:
+                name, count = match.groups()
+                tests.extend([name]*int(count))
+            else:
+                tests.append(test_name)
```
Paul-Ferrell, line 216: Make this work in both direction.  
 - 5*test
 - test*5

Paul-Ferrell: You can drop the re entirely. 
Just check that '*' in test_name, and then do a single split(test_name, '*', 1).

Paul-Ferrell: You also need to add this to the command argument documentation.

Paul-Ferrell (changes requested):

kjeverson:
Still need to drop `<suite_name>.*` support

Paul-Ferrell (approved):

Paul-Ferrell (approved):

PR #401, opened 2021-03-17
depends_pass: True doesn't count skipped test as failures

### LINKED ISSUE #400: `depends_pass: True` doesn't run set
if a set has `depends_pass: True`, that set does not run even if the set it depends on passed

francinelapid:
Fixes #400

### DIFF
--- a/lib/pavilion/series.py
+++ b/lib/pavilion/series.py
@@ -216,7 +216,7 @@ class TestSet:
             try:
                 if test_obj.results['result'] != 'PASS':
                     # treat "SKIPPED" tests as "PASS"
-                    if not test_obj.status.has_state(STATES.SKIPPED):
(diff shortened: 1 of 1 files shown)


lib/pavilion/series.py
```diff
@@ -215,7 +215,9 @@ def is_done(self):
             # check if test passed
             try:
                 if test_obj.results['result'] != 'PASS':
-                    all_tests_passed = False
+                    # treat "SKIPPED" tests as "PASS"
+                    if not test_obj.status.has_state(STATES.SKIPPED):
+                        all_tests_passed = False
```
Paul-Ferrell, line 220: Test objects have a 'skipped' attribute. Use that instead.

Paul-Ferrell (approved):

Issue #334, opened 2020-10-21
Pav Status/Result hangs, Run_Complete into BUILD_CREATED


CalvinDSeamons:
I ran into a very strange bug while testing snow today. This happened on the Yellow front end where there are approximately 11,000 more tests sitting in the working_dir than the Turquoise, just wanted to mention that as that seems to be the only notable difference. 

After launching my tests I ran `watch pav status`. Upon the loading of the status table (which only took 3-5 seconds) a few license-tests had already completed with `PASS` and the rest where all `SCHEDULED`, everything seemed fine. 20ish minutes later after everything had finished my `watch pav status` which updates every 3 seconds showed everything as fine, `PASS`. I quit out (don't ask me why) and ran just `pav status` as to copy the contents out into the ticket. The command hanged, as did `pav result` or any permutation of `pav log build/run $series/$id` ect. I even logged into snow from a different terminal session, loaded pavilion/2.0 and could not access the test run. Upon using the`cat` command i received the following status file from one of the tests that I had observed passing:
```bash
2020-10-21T11:21:08.484894 CREATED Created status file.
2020-10-21T11:21:08.485955 CREATED Test directory and status file created.
2020-10-21T11:21:08.490230 BUILD_CREATED Builder created.
2020-10-21T11:21:08.493727 CREATED Test directory setup complete.
2020-10-21T11:21:16.778993 BUILD_REUSED Test 171aceb2e5e39623 run 13242 reusing build.
2020-10-21T11:21:23.182369 SCHEDULED Test slurm has job ID 3814891.
2020-10-21T11:47:35.252408 PREPPING_RUN Converting run template into run script.
2020-10-21T11:47:35.255769 RUNNING Starting the run script.
2020-10-21T11:47:35.261001 RUNNING Currently running.
2020-10-21T11:47:35.282176 RUN_DONE Test run has completed.
2020-10-21T11:47:35.289546 RESULTS Parsing 6 result types.
2020-10-21T11:47:35.292427 RESULTS Performing 0 result evaluations.
2020-10-21T11:47:35.308790 COMPLETE The test completed with result: PASS
2020-10-21T12:02:03.342975 BUILD_CREATED Builder created.
```
The `PASS` is what I observed inside `watch pav status`. When I exited `watch pav status` the test status changed to `BUILD_CREATED` and was unreachable from `pav status`. 

[...]

PR #402, opened 2021-03-22
Hard Fail with Cancel Changes

kjeverson:
`pav run --hard-fail` on a clean, up to date branch. (Previously PR #372)

### DIFF
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -355,14 +355,32 @@ class TestBuilder:
                 "Source location '{}' points to something unusable."
                 .format(found_src_path))
 
-    def build(self, cancel_event=None):
+    def build(self, cancel_event=None, fail_event=None):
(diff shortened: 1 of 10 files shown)


kjeverson:
I managed to reintroduce the `BUILD_CREATED1 bug (#334 )

Saw that in the test run's `__init__()`, we generate a build/build tracker regardless of state and I don't think we need that if the run is complete. So I added a `if not self.complete` conditional and it seemed to have fixed the issue.

kjeverson:
I tack on an additional status for aborted tests...maybe not the best implementation, but not sure how else it can be done.
Last two lines of an aborted test's status file: 
```
1617739548.948355 ABORTED Build '2a4fd9351095b201' failed under another test.
1617739549.250165 ABORTED Build '2a4fd9351095b201' failed under test 60.
```

PR #406, opened 2021-05-06
added max() and min() expression functions

holmanb:


Paul-Ferrell:
Can you add unit tests for this? 
The correct place is to just add them in test/tests/expression_functions.py.

PR #419, opened 2021-07-27
Parallelize test filter selection.

dmageeLANL:
Encapsulate the main loop over paths in dir_db.select_from to transform and filter the paths in parallel.

Two consequences of this change motivated the other changes.
First, functions declared within other functions can't be pickled.  So move filters.filter_test_run out of the make_test_run_filter function and return it with the same arguments using functools.partial and the make_test_run_filter arguments.
Second, lambdas can't be pickled. So let order_func be a function defined in the top level of the filters module, sort_func, rather than a lambda stored as a value in a dictionary. This eliminates the SERIES_SORT_FUNC and TEST_SORT_FUNC dictionaries and replaces them with a dictionary with keys TEST and SERIES and values that are valid sort keys for these types of dictionaries. The order_func acts the same as previously, but any function that took the SORT_FUNC dicts as inputs now take a string (TEST or SERIES) instead.

### DIFF
--- a/lib/pavilion/filters.py
+++ b/lib/pavilion/filters.py
@@ -243,6 +243,7 @@ def filter_test_run(
     """Determine whether the test run at the given path should be
     included in the set. This function with test_attrs as the sole input is
     returned by make_test_run_filter.
+    
     :param test_attrs: Dict of attributes filtered to determine whether to
     keep or discard test.
     Arguments set by make_test_run_filter.
--- a/lib/pavilion/filters.py
+++ b/lib/pavilion/filters.py
@@ -243,10 +243,9 @@ def filter_test_run(
     """Determine whether the test run at the given path should be
     included in the set. This function with test_attrs as the sole input is
     returned by make_test_run_filter.
-    
+
     :param test_attrs: Dict of attributes filtered to determine whether to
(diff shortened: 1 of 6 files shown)


lib/pavilion/filters.py
```diff
@@ -224,13 +228,55 @@ def add_series_filter_args(arg_parser: argparse.ArgumentParser,
...
 
-    sort_options = (list(sort_functions.keys())
-                    + ['-' + key for key in sort_functions.keys()])
+    sort_options = (sort_functions
+                    + ['-' + key for key in sort_functions])
 
     add_common_filter_args("series", arg_parser, defaults, sort_options)
 
+#  select once so we only make one filter.
+def filter_test_run(test_attrs, cmp, fail, incmp, nm, nt, ot, pss, rslte, ss, sn, usr):
```
Paul-Ferrell, line 239: While I understand the desire for brevity, these arguments need to be longer and more readable, and documented in the docstring.

dmageeLANL: Got it.  Seems like it works now I had some issues with the docstrings.

Paul-Ferrell (approved):

PR #420, opened 2021-07-28
Parallelize results query.

dmageeLANL:
Add file result_utils.py to provide an analogue of status_utils.py for the results command.
The run method from the result.py plugin now calls get_results rather than initializing TestRun.load itself.  Also, get_results returns the test.results dict rather than a TestRun object (which can't be returned by a parallel map invocation because in that case pickle and the TestRun object don't get along.  A result_log field is added to that dict in case show_log is True.  This also simplifies the run command.  The update_results method is not parallel and still requires a full TestRun object so the argument tests is changed to test_ids, a list of ints, which are then initialized in update_results.

Basic profiling numbers for this series of MRs are available in perf.md.
[perf.md](https://github.com/hpc/pavilion2/files/6896228/perf.md)


Paul-Ferrell (dismissed):

Paul-Ferrell:
You're going to have some style check failures. There's instructions in the test/README file on how to set that up.

Paul-Ferrell (approved):

PR #423, opened 2021-09-10
Multi config

Paul-Ferrell:


Paul-Ferrell:
@francinelapid - This is a bit of a big one. 

- You can now have a config file (config.yaml) in each config directory that gives that directory (and its resulting test runs) a label and can customize its working directory. 
  
  - The label only applies to runs, not the actual test names.
  - The default label is 'main'.
  - While test runs know their label, the label is only really important when identifying the test run in the first place.
 - Debugging all this led to the discovery that there were weird import loops amongst all the series stuff.
   - I moved all the series files into their own sub-module.
   - I simplified the concept of a series. Before it was a collection of tests and maybe test sets that also maybe held tests?
     - Now a series is a collection of test sets as defined by a series configuration (and holds the tests created by those test sets).
     - A test_set is an object that wraps a specification for a list of test, including host, modes, etc. 
     - Test sets now internally manage the creation, building, and kickoff of those specified tests.
     - Test Series hold test sets, but test sets don't need to know anything about the series (which was the bad import loop).
 - A few other things got heavily refactored in the process. 
   - The wait command in particular (it was mostly repeated code that's now in cmd_utils).
   - A slew of new test_set tests were added, and some old tests that checked the functionality of the run command (which is now managed by test_sets) were deleted.

bin/pav.py
```diff
@@ -56,35 +55,6 @@ def main():
             color=output.RED)
         sys.exit(-1)
 
-    # Create the basic directories in the working directory and the .pavilion
```
Paul-Ferrell, line 59: This is all done as part of pavilion config setup now.

lib/pavilion/unittest.py
```diff
@@ -74,14 +74,32 @@ def __init__(self, *args, **kwargs):
         """Setup the pav_cfg object, and do other initialization required by
         pavilion."""
 
+        self.pav_cfg = self.make_test_config()
```
Paul-Ferrell, line 77: I just moved this out of \_\_init\_\_ and into its own function.

RELEASE.txt
```diff
@@ -4,6 +4,27 @@ RELEASE=2.3
 
 # Release History
```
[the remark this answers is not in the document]
Paul-Ferrell: Another thing you'll see is that instantiating TestRun objects now has two steps. 

1. Creating the object `TestRun()` - Initializes the object from its config, and creates a test run dir.
2. Saving the object `test_run.save()` - Creates all the files related to the test run and saves them. 

This will become important in future merge requests, as it separates out almost all the IO from getting the test to a state where we can evaluate if it should be skipped.

PR #424, opened 2021-09-14
Dmagee/cache speedup

dmageeLANL:
Fixed bug and excess output.  Added cache to improve wrapped output.


dmageeLANL:
Fixes #426.

dmageeLANL:
It looks like it just timed out making the docs (which I didn't touch). I think the tests will pass if the conditions are normal.

Paul-Ferrell:
I thought I had that function producing a lookup table, but this implicitly does the same.

Paul-Ferrell (approved):

Paul-Ferrell:
I'm not going to merge this until chain of PR's is merged.

PR #425, opened 2021-09-15
Superfast

Paul-Ferrell:
A variety of Pavilion speed enhancements.


Paul-Ferrell:
I'm doing this as a series of smaller pull requests. Note that it's a sub-branch of #423, and won't really make sense for review until after that is merged.

PR #434, opened 2021-09-30
Ticket 431

Paul-Ferrell:


Paul-Ferrell:
Relies on superfast 3, so merge that first.

PR #444, opened 2021-11-04
Config warning fix

dmageeLANL:
Don't have pav_config load empty configs for config.yaml files that don't exist and don't need to exist.  Because the logic around its label key gets wonky and prompts a warning.


dmageeLANL:
OMG finally!

lib/pavilion/config.py
```diff
@@ -33,13 +33,10 @@
 PAV_CONFIG_DIR = os.environ.get('PAV_CONFIG_DIR', None)
 
 if PAV_CONFIG_DIR is not None:
-    PAV_CONFIG_DIR = Path(PAV_CONFIG_DIR)
+    PAV_CONFIG_DIR = Path(PAV_CONFIG_DIR).resolve()
```
Paul-Ferrell: Resolve only works in python 3.5 if the path exists.

Paul-Ferrell (approved):

dmageeLANL:
So.  Is it closed?

Paul-Ferrell:
I was waiting on the unittests to pass and forgot about it.

PR #445, opened 2021-11-08
Fixed error when calling pav status with series id

dmageeLANL:
Calling pav status with series id results in a "test directory does not exist error".  This fixes that bug by climbing up the tree from test path until the path to the test is found in the test_wd/run_dir/test_id.  If the test can't be found it throws a value error.

Previously, the test directory could not be found because the get_tests_by_paths function was hardwired to climb up two directories from the test_path it was given.  When the test_path was a path to a test within a series:

`working_dir/series/<series_id>/<test_id>`.

This meant it looked only in:

`working_dir/series/test_runs/<test_id>`

which doesn't exist.  In this case the string 'test_runs' is TestRun.RUN_DIR.


Paul-Ferrell:
Your solution won't work. 
A test's working directory and the series working directory aren't necessarily the same, so walking up to the series working directory is wrong in some cases. 
What we need to do is resolve the symlink in the series directory to the actual test directory, and go from there. 

I'll make the change in a different MR.

PR #452, opened 2022-01-07
Add generalized IOR tests with xroads-specific subtests.

nicholas-sly:
Adding the IOR test to the examples including subtests for the Crossroads acceptance testing definitions to serve as examples for how some of the more advanced features can be utilized.


examples/pre-packaged/tests/ior.yaml
```diff
@@ -0,0 +1,295 @@
...
+    - size: "2000000"
+      spec: "2M"
+    - size: "4000000"
+      spec: "4M"
+    - size: "8000000"
+      spec: "8M"
+    segcount: '{{ floor( filesize / ( transsize.size * numtasks ) ) }}'
+    mempernode:
+    - percent: '80%'
+      amount: '{{ floor( sched.alloc_max_mem * 0.8 ) }}'
```
Paul-Ferrell, line 253: I'm pretty sure `alloc_max_mem` doesn't exist anymore. There's a `test_min_mem` now though, which is probably more appropriate.

Paul-Ferrell: In fact, there shouldn't be any `alloc_` vars.

nicholas-sly: My favorite part is that `pav show` just throws a pile of errors rather than giving me the names and descriptions of applicable variables.

Paul-Ferrell: I thought it it did that right, but I guess not.

Paul-Ferrell (approved):

PR #456, opened 2022-01-20
Remove -w flag from subprocess calls. Fix mpirun command.

dmageeLANL:
Pavilion itself is getting unavailable nodes in it's node list.  I could have fixed that but it would have required more digging and I think this solution is more parsimonious.  Setting the number of nodes and a list of particular nodes in sbatch/srun/mpirun makes the allocation overdetermined and brittle.  So I commented the -w lines out, you can remove them fully if you agree with my solution.  Also the mpirun command had no -np argument which is necessary.


Paul-Ferrell:
Use `node_state: available` to avoid most of the conflicts. 

Specifying the nodes is required for a bunch of the new scheduler features, but isn't required if those features aren't being used. I'm working on it...

Paul-Ferrell:
Fixed in my latest merge request.

PR #464, opened 2022-02-07
Fixes --sort-by argument.

dmageeLANL:
Before this PR, calling `pav result` will result in a random ordering of output rows.  Calling `pav status` will result in output rows sorted by id regardless of what argument is given to `--sort-by`.

This is because after the tests have been filtered, these commands call `cmd_utils.get_tests_by_paths`, which calls `load_tests` from `test_run/utils.py`, which immediately calls set on the list of tests to remove duplicates but also randomizes their order undoing the sort filter.  This would make the output random for both commands, but the status command calls status_utils in print_status which immediately sorts the tests by their id.  Using another method of removing duplicates (already half implemented by the load_tests function), and removing the sort in print_status resolves the issue.

-------

A final note on the topic of the `--sort-by` flag: using `-` at the start of the sort key to denote descending order is a bad idea.  Argparse uses `-` to recognize flags so passing `--sort-by -id` throws an error:
```
pav results: error: argument --sort-by: expected one argument
```

The key must be passed to the flag with `=` which is not denoted in the help string.  But this is confusing as the = is only needed for this one flag and only when passing a particular set of keys.  I would suggest just changing the logic to use `+` (or `<` or maybe `~`?) to denote descending order.

### DIFF
--- a/lib/pavilion/status_utils.py
+++ b/lib/pavilion/status_utils.py
@@ -133,8 +133,6 @@ def print_status(statuses, outfile, note=False, series=False, json=False):
 :rtype: int
 """
 
-    statuses.sort(key=lambda v: v.get('test_id'))
-
     if json:
         json_data = {'statuses': statuses}
         output.json_dump(json_data, outfile)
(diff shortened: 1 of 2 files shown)


Paul-Ferrell (approved):
I realized I might have broke ordering, and but forgot where I did it, and have been casually looking for this for days.

PR #506, opened 2022-05-27
Status improvements

Paul-Ferrell:


lib/pavilion/cmd_utils.py
```diff
@@ -136,7 +136,7 @@ def arg_filtered_series(pav_cfg: config.PavConfig, args: argparse.Namespace,
             output.fprint(verbose, "Using default search filters: The current system, user, and "
                                    "newer_than 1 day ago.", color=output.CYAN)
             args.user = utils.get_login()
```
Paul-Ferrell, line 138: This and the next change fix a bug in listing series - the series filter creator took the wrong argument types.

lib/pavilion/jobs.py
```diff
@@ -104,7 +104,7 @@ def __str__(self):
         parts = []
         if self.info is not None:
             for key in sorted(self.info.keys()):
```
Paul-Ferrell, line 106: This was why the job_id's were funky.

lib/pavilion/schedulers/advanced.py
```diff
@@ -10,7 +10,7 @@
 from pavilion.status_file import STATES
 from pavilion.test_run import TestRun
 from pavilion.types import NodeInfo, Nodes, NodeList, NodeSet, NodeRange
-from .config import validate_config, AVAILABLE, BACKFILL
```
Paul-Ferrell, line 13: calc_node_range was a static method (basically a function tacked on to a class, rather than a full method) attached to the scheduler plugin class. I moved it to the scheduler config module as a regular function. It's a better fit there anyway, and that let me use it to get the 'requested_nodes' as a scheduler variable consistent with how it was calculated elsewhere.

PR #510, opened 2022-06-14
Added a name filter argument for `pav show tests`

smehta99:
- Added a name filter argument to filter show tests use `pav show tests <filter>`
- Added path for the test & suite files. Use `pav show tests/suites -p/--path` for path.

### DIFF
--- a/lib/pavilion/commands/show.py
+++ b/lib/pavilion/commands/show.py
@@ -3,6 +3,7 @@
 import argparse
 import errno
 import pprint
+import fnmatch
 from typing import Union
 
 import pavilion.types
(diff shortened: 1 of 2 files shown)


lib/pavilion/commands/show.py
```diff
@@ -316,13 +317,27 @@ def _setup_arguments(self, parser):
...
         )
+        tests.add_argument(
+            'name_filter', type=str, nargs='?', default='',
+            help="Filter tests."
+        )
+        tests.add_argument(
+            '--path', '-p',
+            action='store_true', default=False,
+            help='Print the path for each suite file.'
+        )
```
Paul-Ferrell, line 340: We don't need this twice.

Paul-Ferrell (changes requested):

Paul-Ferrell (approved):

PR #511, opened 2022-06-22
Fixed specifying slurm constraints/features.

Paul-Ferrell:


Paul-Ferrell:
Style failure was due to the fact that distutils is now deprecated. Will fix in a separate PR.

PR #524, opened 2022-07-23
Add support for nodenames foo-[10-20]-[01-02].

aebruno:
Our node names look this this `cpn-m11-16`. When running pav we get the following error:

```
Unknown error running command _run: Invalid Node List: 'cpn-m11-16'. Syntax
error in item 'cpn-m11-16'. Node lists components be a hostname or hostname
prefix followed by a range of node numbers. Ex: foo003,foo0[10-20],foo[103-104],
foo[10,12-14].
```

This PR adjusts the node regex to support this naming convention.

### DIFF
--- a/lib/pavilion/schedulers/plugins/slurm.py
+++ b/lib/pavilion/schedulers/plugins/slurm.py
@@ -195,9 +195,9 @@ class Slurm(SchedulerPluginAdvanced):
 
     NODE_SEQ_REGEX_STR = (
         # The characters in a valid hostname.
-        r'[a-zA-Z][a-zA-Z_-]*\d*'
+        r'[a-zA-Z-][a-zA-Z0-9_-]*\d*'
         # A numeric range of nodes in square brackets.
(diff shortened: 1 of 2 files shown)


aebruno:
I was unable to get this test to pass: 

```
======================================================================
FAIL: test_node_list_parsing (slurm_tests.SlurmTests)
Make sure the node list regex matches what it's supposed to.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/aeb/projects/ccr/pavilion2/test/tests/slurm_tests.py", line 129, in test_node_list_parsing
    slurm.parse_node_list(ex)
AssertionError: ValueError not raised : Did not throw error for Trailing characters
```

Is there any reason why we can't support these hostnames? `n03d`

Paul-Ferrell (approved):

Paul-Ferrell:
There's a few corner cases that this exposed that the unittests caught. I've got a followup PR to fix those after this merge.

aebruno:
@Paul-Ferrell thanks so much for getting this merged. Just tested out #524 and #525 on our cluster and all works well. Thanks again!

PR #545, opened 2022-09-20
Add/flux framework scheduler

vsoch:
This is the first draft to add the Flux Scheduler to Pavilion2! The Flux way of running jobs isn't nicely matched to slurm so some of these interactions were harder to figure out, but this should be a good start for us to discuss. Thanks for all the great help @pflarr !

### DIFF
--- a/lib/pavilion/schedulers/plugins/flux.py
+++ b/lib/pavilion/schedulers/plugins/flux.py
@@ -1,0 +1,267 @@
+# pylint: disable=too-many-lines
+"""The Flux Framework Scheduler Plugin."""
+
+import os
+import time
+from typing import List, Union, Any, Tuple
+
+import yaml_config as yc
+from pavilion import sys_vars
+from pavilion.jobs import Job, JobInfo
+from pavilion.status_file import STATES, TestStatusInfo
+from pavilion.types import NodeInfo, NodeList
+from ..advanced import SchedulerPluginAdvanced
+from ..config import validate_list
+
+
+# Just import flux once
+try:
+    import flux
+    import flux.job
+    import flux.resource
+    from flux.job import JobspecV1
+except ImportError:
+    flux = None
+
+flux_states = [
+    "DEPEND",
+    "SCHED",
+    "RUN",
+    "CLEANUP",
+    "INACTIVE",
+]
+
+
+class Flux(SchedulerPluginAdvanced):
+    """
+    Schedule tests with Flux!
+    """
+
+    def __init__(self):
+        super().__init__("flux", "Schedules tests via the Flux Framework scheduler.")
+
+    def _get_config_elems(self):
+        elems = [
+            yc.ListElem(
+                name="avail_states",
+                sub_elem=yc.StrElem(),
+                help_text="When looking for immediately available "
+                "nodes, they must be in one of these "
+                "states.",
+            ),
+            yc.ListElem(
+                name="up_states",
+                sub_elem=yc.StrElem(),
+                help_text="When looking for nodes that could be  "
+                "allocated, they must be in one of these "
+                "states.",
+            ),
+        ]
+
+        defaults = {
+            "up_states": flux_states,
+            "avail_states": flux_states,
+        }
+
+        validators = {
+            "up_states": validate_list,
+            "avail_states": validate_list,
+        }
+
+        return elems, validators, defaults
+
+    def _get_alloc_nodes(self, job) -> NodeList:
+        """
+        Get the list of allocated nodes.
+        """
+        listing = flux.job.list.JobList(flux.Flux(), ids=[job.info["jobid"]])
+        jobs = listing.jobs()
+        if not jobs:
+            return []
+        nodes = jobs[0]._nodelist
+        if not isinstance(nodes, list):
+            nodes = [nodes]
+        return nodes
+
+    def _get_raw_node_data(self, sched_config) -> Tuple[Union[List[Any], None], Any]:
+        """
+        Get a flux resource list
+        """
+
+        rpc = flux.resource.list.resource_list(flux.Flux())
+        listing = rpc.get()
+
+        nodelist = [str(node) for node in listing.free.nodelist]
+        extra = {"nodes_listing": listing}
+        return nodelist, extra
+
+    def _transform_raw_node_data(self, sched_config, node_data, extra) -> NodeInfo:
+        """
+        Translate the gathered data into a NodeInfo dict.
+        """
+        listing = extra["nodes_listing"]
+        del extra["nodes_listing"]
+        node_info = NodeInfo({})
+        node_info["name"] = node_data
+        node_info["cpus"] = listing.free.ncores
+        node_info["up"] = node_data in list(listing.up.nodelist)
+        node_info["available"] = node_data in list(listing.free.nodelist)
+        return node_info
+
+    def _available(self) -> bool:
+        """
+        Ensure we can import and talk to flux.
+        """
+        return flux is not None
+
+    def _kickoff(self, pav_cfg, job: Job, sched_config: dict) -> JobInfo:
+        """
+        Submit the kick off script using Flux.
+        """
+
+        output = job.sched_log.as_posix()
+        error = output.replace(".log", ".err")
+
+        # Generate the flux job
+        # Assume the filename includes a hashbang
+        # flux does not support mem_mb, disk_mb
+        fluxjob = JobspecV1.from_command(
+            command=["/bin/bash", job.kickoff_path.as_posix()],
+            num_tasks=sched_config["tasks_per_node"],
+        )
+
+        # Min time is one minute
+        limit = sched_config["time_limit"] or 0
+        limit = limit * sched_config["time_limit"] * 60
+
+        # A duration of zero (the default) means unlimited
+        fluxjob.duration = limit
+        fluxjob.stdout = output
+        fluxjob.stderr = error
+
+        # This doesn't seem to be used?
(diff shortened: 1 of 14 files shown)


lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+flux_states = [
+    "DEPEND",
+    "SCHED",
+    "RUN",
+    "CLEANUP",
+    "COMPLETED",
+    "FAILED",
+    "CANCELED",
+    "TIMEOUT",
+]
```
Paul-Ferrell, line 35: The equivalent in the Slurm scheduler was about figuring out the state of each node, but these look like job states. If that's what you're using them for, that's fine.

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+        }
+
+        return elems, validators, defaults
+
+    @classmethod
+    def parse_node_list(cls, node_list) -> NodeList:
+        """
+        Convert a list of strings into a list of nodes
+        """
+        return NodeList(node_list)
```
Paul-Ferrell, line 81: You can just get rid of this, and do this inline. A lot of these functions are Slurm specific. Check the Advanced and base scheduler classes for the methods that must be overridden (they'll all raise NotImplementedError).

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+        Get a flux resource list
+        """
+
+        rpc = flux.resource.list.resource_list(flux.Flux())
+        listing = rpc.get()
+
+        nodelist = [str(node) for node in listing.free.nodelist]
+        # Flux doesn't currently support reservations
+        extra = {"reservations": {}, "nodes_listing": listing}
+        return nodelist, extra
```
Paul-Ferrell, line 107: The data this returns is entirely dependent on the Scheduler. No need to still return anything about reservations.

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+    def _transform_raw_node_data(self, sched_config, node_data, extra) -> NodeInfo:
+        """
+        Translate the gathered data into a NodeInfo dict.
+        """
+        listing = extra["nodes_listing"]
+        del extra["nodes_listing"]
+        node_info = NodeInfo({})
+        node_info["name"] = node_data
+        node_info["cpus"] = listing.free.ncores
+        node_info["states"] = None
```
Paul-Ferrell, line 118: See the superclass for the list of required keys. Anything else is scheduler specific, including 'states'.

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+        node_info["available"] = node_data
+        return node_info
+
+    def _filter_custom(
+        self, sched_config: dict, node_name: str, node: NodeInfo
+    ) -> Union[str, None]:
+        """
+        Filter nodes by features. Flux doesn't have features.
+        """
+        return None
```
Paul-Ferrell, line 129: The default does nothing, so you can scrap this.

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def _available(self) -> bool:
+        """
+        Ensure we can import and talk to flux.
+        """
+        return flux is not None
+
```
Paul-Ferrell, line 136: Everything from here down looks good.

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+        Translate the gathered data into a NodeInfo dict.
+        """
+        listing = extra["nodes_listing"]
+        del extra["nodes_listing"]
+        node_info = NodeInfo({})
+        node_info["name"] = node_data
+        node_info["cpus"] = listing.free.ncores
+        node_info["states"] = None
+        node_info["up"] = node_data
+        node_info["available"] = node_data
```
Paul-Ferrell, line 120: These have a very important and specific meaning (from the parent method doc string):

>        # Node Status - (required)
>
>        - up (bool) - Whether the node is up (allocatable).
>        - available (bool) - Whether the node is allocatable and unallocated.

lib/pavilion/schedulers/plugins/flux.py
vsoch: ah gotcha.

lib/pavilion/schedulers/plugins/flux.py
vsoch: Where is that? I did a grep and only found a test file dummy.py

Paul-Ferrell: schedulers.advanced.py

lib/pavilion/schedulers/plugins/flux.py
vsoch: So the sets overlap?

vsoch: okay we actually have both free and up lists to match nicely to these two!

lib/pavilion/schedulers/plugins/flux.py
vsoch: Thanks! Okay so the list is:

- name: we have this
- status (up/available): now are correctly specified (bool)
- cpus: we have this
- mem: I didn't see this information with the listing

And these are slurm specific I think.

- partitions (List) - The cluster partitions on which the node resides.
- reservations (List) - List of reservations to which the node belongs.
- features (List[str]) - A list of feature tags that differentiate nodes,

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,274 @@
...
+        if job_info["sys_name"] != sys_name:
+            return TestStatusInfo(
+                STATES.SCHEDULED,
+                "Job started on a different cluster ({}).".format(sys_name),
+            )
+
+        listing = flux.job.list.JobList(flux.Flux(), ids=[job_info["jobid"]])
+        jobs = listing.jobs()
+
+        if not listing:
```
grondo, line 185: Should this be `if not jobs`?

grondo (commented):
Just adding some comments on use of the Flux APIs after a cursory inspection.

lib/pavilion/schedulers/plugins/flux.py
Paul-Ferrell: partitions and reservations are general, but optional.

lib/pavilion/schedulers/plugins/flux.py
Paul-Ferrell: Style note - Pavilion has a separate job object, so we should probably be disambiguating them.  Maybe call them `flux_jobs`?

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,291 @@
...
+                note="Could not find job {}, must have finished".format(job_info["id"]),
+                when=time.time(),
+            )
+
+        # State list is here
+        # https://flux-framework.readthedocs.io/projects/flux-rfc/en/latest/spec_21.html#state-diagram
+        # DEPEND
+        job = jobs[0]
+
+        if job.state == "COMPLETED":
```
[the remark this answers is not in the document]
vsoch: ah ok, so I should be checking status here!

lib/pavilion/schedulers/plugins/flux.py
vsoch: yes it should! I renamed / debugged this a bunch of times and forgot to update the name here, should be fixed now.

jameshcorbett (commented):
One insignificant comment below.

lib/pavilion/schedulers/plugins/flux.py
```diff
@@ -0,0 +1,267 @@
...
+
+        # This doesn't seem to be used?
+        fluxjob.cwd = str(job.tests_path)
+        fluxjob.environment = dict(os.environ)
+
+        # This submits without waiting
+        flux_future = flux.job.submit_async(flux.Flux(), fluxjob)
+
+        # This blocks until the RPC is complete so we will get an ID
+        jobid = flux_future.get_id()
```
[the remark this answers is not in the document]
vsoch: ah! so submit_async doesn't wait for the id, but I'm asking it to wait with get_id? So if I just use submit that does two in one?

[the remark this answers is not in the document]
jameshcorbett: Yeah. Sorry for not answering earlier.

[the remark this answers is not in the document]
jameshcorbett: [The code for `submit` is actually just a wrapper around submit_async](https://github.com/flux-framework/flux-core/blob/99ecf1da6ebcacf4e53ced93f0088c039de1d12e/src/bindings/python/flux/job/submit.py#L103)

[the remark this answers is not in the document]
vsoch: no worries! I think I realized this when I was working on the asyncio stuff - apologies for asking I should have just looked, still familiarizing with the python code.

Paul-Ferrell:
Flux scheduler finished an merged in #674

vsoch:
Nice work @Paul-Ferrell so happy to see that merged! :partying_face:

PR #548, opened 2022-09-21
Improve results keys

dmageeLANL:
Improvements to the `pav results` command:

- Sorts key list when `--list-keys` is given.
- Includes var dict in results.json so variables can be given as keys on command line.
- Uses default if invalid `--sort-by` key is given rather than failing.
- Sorts on result.json key if given with `-sort-by` flag.

I'll write some docstrings and such, I just wanted to put it out there in the meantime.


Paul-Ferrell (commented):
I don't see any issues here other than a lack of comments. I need to pull this down and do hands-on testing though.

dmageeLANL:
Comments and docstrings applied.

Two main notes: 
1. I stopped checking if the sort by key was in the fields list and removed fields from the argument list. Sorting by a hidden key should be allowed.
2. I debated what the dummy value should be in the sort but ultimately settled on the lowest value, that is, by default the tests without the key will be at the top.  That struck me as wrong at first, the sorted bunch should come first.  But then I thought it was more important to guarantee that the sorted group (in default ascending order) was immediately visible and so at the bottom of the list.

Paul-Ferrell (approved):

dmageeLANL:
It bothered me that it isn't easier to get the name of the current series so this commit has the results table and keylist print the series that is being displayed or the series/tests requested on the command line.

Paul-Ferrell (approved):

PR #563, opened 2022-10-19
Doc cleanup

Paul-Ferrell:


Paul-Ferrell:
@frakeith - This removes the entirety of the sphinx theme from the repository, and includes it as an outside module instead. It also cleans up some syntax issues in the documentation.

francinelapid (approved):

PR #564, opened 2022-10-19
allow row keys to start with a number

francinelapid:


Paul-Ferrell (dismissed):

Paul-Ferrell:
The expected results for the table_result_parser will need to be updated.

Paul-Ferrell (approved):

PR #570, opened 2022-10-25
Added -f and --follow option to pav log command.

smehta99:
Usage:  `pav log -f/--follow <log_type> <test_id>`

example: `pav log -f build 72`


smehta99:
There is a bug where where build hash log doesn't change from the previous run and ends up not finding the new hash. I will fix it when I return.

lib/pavilion/commands/log.py
```diff
@@ -129,9 +137,58 @@ def run(self, pav_cfg, args):
...
+                                    time.sleep(.5)
+
+                    with file_name.open() as file:
+                        file.seek(0,2)
+                        while True:
+                            data = file.read()
+                            if not data:
+                                time.sleep(.5)
+                            else:
+                                output.fprint(self.outfile, data, end='', flush=True)
```
Paul-Ferrell, line 182: Instead of jumping through all these possibilities, use a list of possible filename and just check each of them in turn.

Paul-Ferrell (changes requested):

lib/pavilion/commands/log.py
```diff
@@ -98,6 +110,22 @@ def _setup_arguments(self, parser):
         'series': 'series.out'
     }
 
+    def print_log(self, current_position, file):
```
Paul-Ferrell, line 113: Add docstring

Paul-Ferrell: and arg types

Paul-Ferrell (changes requested):

lib/pavilion/commands/log.py
```diff
@@ -128,6 +165,29 @@ def run(self, pav_cfg, args):
...
 
+        if args.follow:
+            position = 0
+            while True:
+                if cmd_name == 'build':
+                    build_log = test.builder.log_path
+                    tmp_build_log = test.builder.tmp_log_path
+                    for build_log_path in [file_name, build_log, tmp_build_log]:
+                        if build_log_path.exists():
+                            with build_log_path.open() as file:
```
Paul-Ferrell, line 176: Race condition

Paul-Ferrell (approved):

lib/pavilion/commands/log.py
```diff
@@ -128,23 +153,54 @@ def run(self, pav_cfg, args):
...
+                            with file_path.open() as file:
+
+                                # Prints the last n lines.
+                                if args.tail:
+                                    tail = file.readlines()[-int(args.tail):]
+                                    for line in tail:
+                                        output.fprint(self.outfile, line)
+
+                                # Follows the execution of the code.
+                                elif args.follow:
```
Paul-Ferrell, line 181: These shouldn't be mutually exclusive. What if I want to tail and follow?

PR #573, opened 2022-10-27
Tableparser

jennfshr:
This fixes an issue we have with the table parser in IMB.

It removes the "row_" prefix from row labels, and removes the constraint that row labels must not start with a number.


Paul-Ferrell (approved):

Paul-Ferrell:
Francine had already fixed this.

PR #574, opened 2022-10-27
Fixed the exception message bug.

Paul-Ferrell:
Pavilion errors take a message and then an (optional) prior exception as arguments. These unintentionally passed the prior exception to `.format()`.


smehta99 (approved):
lgtm.

PR #594, opened 2022-12-21
Added store_min, store_med, and store_max to result parser.  

smehta99:


smehta99:
Pulled off of my smehta-follow branch instead of master. Whoops. Let me fix that.

PR #595, opened 2022-12-21
Added store_min, store_med, store_max to result parser

smehta99:


lib/pavilion/result/options.py
```diff
@@ -100,11 +103,72 @@ def action_store_sum(raw_val):
...
+    if isinstance(raw_val, list):
+        val_min = None
+        for val in raw_val:
+            val = auto_type_convert(val)
+            if isinstance(val, (int, float)):
+                if val_min is None:
+                    val_min = val
+                else:
+                    if val_min > val:
+                        val_min = val
```
smehta99, line 118: Would it be faster to create a list first, then sort that list and then get the min/max?

Paul-Ferrell (approved):

PR #605, opened 2023-01-19
Added a system variable for processor count

smehta99:
Added the `{{host_cpus}}` system variable for parallel builds. 
Ex: `./config -j {{host_cpus // 2}}`

### DIFF
--- a/lib/pavilion/sys_vars/cpu_count.py
+++ b/lib/pavilion/sys_vars/cpu_count.py
@@ -2,11 +2,11 @@ import subprocess
 from .base_classes import SystemPlugin
 
 
-class SystemProcessorCount(SystemPlugin):
+class CPUCount(SystemPlugin):
 
     def __init__(self):
         super().__init__(
-            name='sys_processor_count',
+            name='cpu_count',
             description="The system processor count.",
             priority=self.PRIO_CORE)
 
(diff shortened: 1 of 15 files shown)


lib/pavilion/sys_vars/cpu_count.py
```diff
@@ -0,0 +1,17 @@
...
+        super().__init__(
+            name='cpu_count',
+            description="The system processor count.",
+            priority=self.PRIO_CORE)
+
+    def _get( self):
+        """Base method for determining the system processor count."""
+
+        name = subprocess.check_output(['grep', '-c', '^processor', '/proc/cpuinfo'])
+        return name.strip().decode('UTF-8')
```
Paul-Ferrell, line 17: I'd make the regex more restrictive, to prevent false positive matches `'^processor *:'`.

lib/pavilion/sys_vars/cpu_count.py
```diff
@@ -0,0 +1,17 @@
+import subprocess
+from .base_classes import SystemPlugin
+
+
+class CPUCount(SystemPlugin):
+
+    def __init__(self):
+        super().__init__(
+            name='cpu_count',
```
Paul-Ferrell, line 9: I looked at the existing variable names. `host_cpus` seems to be the most appropriate name.

Paul-Ferrell (approved):

PR #606, opened 2023-01-23
Allow numberless nodes

Paul-Ferrell:


Paul-Ferrell:
Can you check that this works on fog?  We won't have numberless nodes, but just make sure tests still run fine.

frakeith (dismissed):

PR #608, opened 2023-01-23
adding ability to run specific permutations

frakeith:


lib/pavilion/series/test_set.py
```diff
@@ -195,6 +203,19 @@ def make(self, build_only=False, rebuild=False, local_builds_only=False,
...
+                cfg_name = '.'.join([
+                    test_configs[cfg_index].config['suite'],
+                    test_configs[cfg_index].config['name'],
+                    test_configs[cfg_index].config['subtitle']])
+
+                if not fnmatch(cfg_name, pattern):
+                    self._permute_indexes.append(test_configs[cfg_index])
+
+        for pindexes in self._permute_indexes:
+            test_configs.remove(pindexes)
```
Paul-Ferrell, line 217: Given requests for tests 'a.b.*, c.d', the resulting test list would only include the 'a.b.*' tests. 

It's a bit worse than that though.

Let's say the user requests: 'a.b.* c.d a.b.3', given that 'a.b' has permutations 1,2, and 3 and 'c.d' has no permutations

I would expect to get five tests back: 'a.b.1', 'a.b.2', 'a.b.3', 'c.d', and 'a.b.3'.

Paul-Ferrell: To handle that, you really need to do the filtering on a per-request basis. 

Which leads me to a bigger issue - While you _CAN_ put this code in TestSet, I don't think it belongs there. 

It's the job of the Resolver.load() to turn a list of test names into a list of test objects (well, ProtoTest objects). Given the problem I illustrated in the previous comment, I actually think it _HAS_ to go there. By the time the test requests have run through Resolver.load, they're disassociated from the ProtoTest objects that are produced (which you need to filter).

Now, load() doesn't keep up that association, but you have everything you need in that function to do so. That's not the case out in `TestSet.make()`.

lib/pavilion/resolver.py
```diff
@@ -180,7 +180,10 @@ def filter_resolved_tests(self, resolved_tests, filters) -> List[ProtoTest]:
         filtered_tests = []
 
         for test in resolved_tests:
-            tests_dict[test[0]['suite'] + '.' + test[0]['name'] + '.' + test[0]['subtitle']] = test
+            if test[0]['subtitle']:
+                tests_dict[test[0]['suite'] + '.' + test[0]['name'] + '.' + test[0]['subtitle']] = test
+            else:
+                filtered_tests.append(ProtoTest(test[0], test[1].__deepcopy__()))
```
Paul-Ferrell, line 186: Deep copying is what you want, but that's not how you do it.  Use `copy.deepcopy()`, or check that object's built-in `.copy()` function to see if it deep copies by default.

lib/pavilion/resolver.py
```diff
@@ -354,7 +398,33 @@ def load(self, tests: List[str], host: str = None,
...
+        for test in tests:
+            scalar_str = ''
+            scalar_int = 1
+            if fnmatch(test, "*[!.-][*]?*"):
+                test, scalar_str = test.rsplit("*", 1)
+
+                try:
+                    scalar_int = int(scalar_str)
+                except ValueError as err:
+                    raise TestConfigError("Invalid repeat notation".format(err))
```
Paul-Ferrell, line 413: I know I said `sometest*8` was a valid way to ask for tests, but that implies that it's already implemented somewhere in the code. If you can't find where that implementation is, then ask for help. 
_DON'T_ try to re-implement it or handle it yourself without understanding how it works. 

In this case, I was wrong - We used to support that, but I forgot we changed how we did it and it's not really an option anymore. So this whole chunk should go away.

lib/pavilion/resolver.py
```diff
@@ -439,12 +509,14 @@ def load(self, tests: List[str], host: str = None,
                     except TimeoutError:
                         pass
 
-        if outfile:
+        if outfile and not permuted_tests:
```
Paul-Ferrell, line 512: The prints during test resolution are printed with `\r`, so the next percentage printed will overwrite the last. This prints a newline so the next printed thing doesn't overwrite that. 

So this is needed when `outfile and len(permuted_test) > 1`.

lib/pavilion/resolver.py
```diff
@@ -166,6 +167,49 @@ def check_variable_consistency(self, raw_tests: List[Dict]):
                                 .format(test_name, test_suite, var_key, first_value_keys, i + 1,
                                         value_keys))
 
+    def filter_resolved_tests(self, resolved_tests, filters, outfile: IO[str] = None) -> List[ProtoTest]:
```
Paul-Ferrell, line 170: I would rename this `select_permutations`, which makes it more clear what's actually happening.

lib/pavilion/resolver.py
Paul-Ferrell: Ok, I'm wrong again. I finally found where the multiplication takes place, and it's in `load_raw_configs`.  
Which is where you should be handling all this too.

PR #609, opened 2023-01-26
Now allows usage of the same nodes if share_allocation is set to max

smehta99:
```YAML
base:
  schedule:
    share_allocation: max
```


lib/pavilion/schedulers/advanced.py
```diff
@@ -417,9 +417,12 @@ def _schedule_chunk(self, pav_cfg, chunk: NodeSet, tests: List[TestRun],
...
         for acq_opts, tests in share_groups.items():
             chunking_enabled = sched_configs[tests[0].full_id]['chunking']['size'] not in (0, None)
+            # If the user really wants to use the same node even if other nodes are available,
+            # setting share_allocation to max will allow that.
+            use_same_nodes = True if sched_config['share_allocation'] == 'max' else False
             _, max_nodes = node_range = acq_opts[0]
             # If chunking is used, schedule all the tests together.
-            if chunking_enabled:
+            if chunking_enabled or use_same_nodes:
                 self._schedule_shared_chunk(pav_cfg, tests, node_range, sched_configs, chunk)
```
Paul-Ferrell, line 426: Chunking isn't enabled here, so you're calling the wrong function. Doing this will force the job to run on very specific nodes, even though that wasn't asked for.

Paul-Ferrell: Once you fix this, the main thing you're lacking is a test for the new feature in the unit tests.

smehta99: I am not sure what you mean here.
From what I understand, chunks are basically the entire `node_list` if it's not specified. If it is then it will use a smaller portion of the `node_list` specified by the user. 
```Python
328             if chunk_size in (None, 0) or chunk_size > len(node_list):
329                 chunk_size = len(node_list)
330 
331             chunks_id = (node_list_id, chunk_size, node_select, chunk_extra)
332 
333             chunks = self._chunks[chunks_id]
```
Then it goes through a loop to find the least used chunks and then sets the chunk with the list of nodes. Then `_schedule_chuck` is called  which then schedules it based on `share_allocation` option. 

So my logic was by the time it gets to bin separation process, the size of `chunk` is either the specified chuck size the user asked for, or all the nodes available.

Paul-Ferrell: Those comments should be more clear. These groups don't directly correlate to the `_schedule_*` functions. Tests in group 1 (which include all of the tests you're dealing with, may end up going to any of the three `_schedule_*` functions. 

All tests with flexible (non-chunking) scheduling go to `_schedule_flex_tests`.
All chunked tests that don't share an allocation (whether because they're unique or because it's disabled) go to `_schedule_indi_chunk`. 
All chunked tests that do share an allocation got to `_schedule_shared_chunk`.

Tests in the `share_groups` may end up going to any of these three.

Paul-Ferrell (changes requested):

Paul-Ferrell (approved):

PR #612, opened 2023-02-04
List extend

Paul-Ferrell:
This changes the base yaml_config library to allow for extending lists across the board. The new syntax allows for extending any list in a Pavilion config _at a new config layer_ - IE in inherited tests, or modes. 

For example:

```yaml 

base:
  run:
    modules: ['foo']

second:
  inherits_from: base
  run:
    modules+: ['bar']
```
Would result in a modules list with both foo and bar. This syntax always extends the list. There's no way prepend. 
This is independent from how default variables and extending lists of variables works (that's way more complicated).


dmageeLANL:
Merge in my commit to this branch, PR #614 , to pass the style tests :crossed_fingers: and complete the unit tests.  I really want to see if it works.

francinelapid (approved):
Looked over code and tested with rudimentary tests.

PR #649, opened 2023-05-02
Allow nodes

Paul-Ferrell:
Code review checklist:


Paul-Ferrell:
This adds an 'across_nodes' option to the test `schedule` section, allowing you to specify a range of nodes the test can run on. All chunking happens within the given nodes. Non-flexible scheduling is always used, because we have to tell the scheduler specifically what nodes to run on. 

This also fixes a bug I found in `schedule.include_nodes` - it simply didn't work with chunking. Now every chunk always includes those nodes.

PR #653, opened 2023-05-31
Ignore errors

Paul-Ferrell:
This PR allows series to ignore errors in the build and kickoff step. 

Code review checklist:


Paul-Ferrell:
@lapid

Here's the high level view of what's going on. 

- Changed test series so that they can handle errors in the test creation/build/kickoff steps without dying. 
- Test creation is now ~~iterative~~ generative in the test config resolver
  - For every request made for a test (ie 'supermagic'), it gets expanded to actual tests mostly individually.
  - Those are returned via a generator with `yield` statements. (The functions are called `*_iter` because I got iterators and generators confused). 
  - Tests are yielded by the resolver `simultaneous` series setting, which defaults to extremely large. 
  - Test sets work similarly - the make_iter() method now returns tests in batches. 
  - Those batches may reduce in size as builds fail or tests are skipped.
- While all this is going on, any errors are caught and just dumped into lists of errors that are reported by the test_set.
- Most of this reporting goes to the series status file. Depending on the verbosity settings, it may get printed. 
- Verbosity and the `outfile` are now instance attributes and passed to `__init__`. 
The big unknown on this one is how well it works in practice.  Especially with regards to error reporting and messaging. All of this effects doing `pav series run` of course, but it also effects `pav run` as well.

Paul-Ferrell:
Oh, and another big change is that `resolver.py` has been split into `resolver/*`. 
Much of what it did has been moved into `RawProtoTest` (completely unresolved tests) and `ProtoTest` (mostly resolved tests) objects. The TestConfigResolver handles all the system/config interfacing, then all the internals of test resolution are handled by those new objects.

francinelapid (approved):
Everything seems to be working as expected.

PR #671, opened 2023-07-31
Add/flux framework scheduler

nicholas-sly:
Code review checklist:

- [nope] Code is generally sensical and well commented
- [absolutely not] Variable/function names all telegraph their purpose and contents
- [who does this] Functions/classes have useful doc strings
- [python doesn't need this] Function arguments are typed
- [definitely not] Added/modified unit tests to cover changes.
- [code is self-documenting] New features have documentation added to the docs.
- [only forward compatibility] New features and backwards compatibility breaks are noted in the RELEASE.md

Add the flux scheduler plugin.  Tested on LLNL flux-scheduled machines.  Lots of ancillary changes after merging in changes from master.  Will need help determining what are appropriate or should be reverted to match master.

Flux compatibility tries to merge the ethos of Pavilion using scripts to ensure reproducibility with the available python API for Flux.  This is embodied in the allocation method that generates a batch script, but uses the text from that script with the allocation API function to actually submit the request to Flux.


Paul-Ferrell:
Replaced by #674

PR #752, opened 2024-03-01
fix for issue with source_download set to latest

### LINKED ISSUE #711: `source_url` and `source_download` do not work as expected
```
        source_url: <url>
        source_download: missing
```

does not end up downloading the source 

Also, when `source_download` is set to `latest`, I get 
```
...
Could not create builder for test stream.base (run 11797): Could not retrieve source from the given url ...
...
```
but a temp file with what seems like the correct content gets downloaded into the `test_src` directory.

debonisd:
Code review checklist:


lib/pavilion/builder.py
```diff
@@ -319,7 +319,7 @@ def _update_src(self):
 
         if (src_url is not None
                 and ((src_download == 'missing' and found_src_path is None)
-                     or src_download == 'latest')):
```
Paul-Ferrell, line 322: You've just changed the behavior of 'latest' such that it is now doing the same as 'missing'.  That's not the problem.  It seems that the files are getting downloaded, but they're not getting copied into their expected destination where Pavilion can find them.

Paul-Ferrell (changes requested):

Paul-Ferrell:
This is still an empty merge request...

Paul-Ferrell (dismissed):

PR #757, opened 2024-03-19
adding more build options

### LINKED ISSUE #745: Add build options for large external builds
This requires some discussion, but the jist is:

We need to point at builds that we don't actually want to copy. 
I'm thinking this should have two components:
1. An option that says - hey, the build is over there, outside of Pavilion. Trust that it works.
2. An option that says - when you do the symlink copy, only go to depth X. At X - even directories should be symlinks to the original build (the current setup creates real directories with symlink to files in most cases).

francinelapid:
Code review checklist:

Closes #745


lib/pavilion/builder.py
```diff
@@ -666,18 +667,23 @@ def _setup_build_dir(self, dest, tracker: BuildTracker):
...
                       .format(src_path, dest)))
 
-            utils.copytree(
-                src_path.as_posix(),
-                dest.as_posix(),
-                copy_function=shutil.copyfile,
-                copystat=utils.make_umask_filtered_copystat(umask),
-                symlinks=True)
+            source_copy = self._config.get('source_copy')
+            if source_copy.lower() == 'true':
```
Paul-Ferrell, line 678: You can use the utils.py `str_bool` here.

Paul-Ferrell (dismissed):

lib/pavilion/utils.py
```diff
@@ -193,6 +195,18 @@ def path_is_external(path: Path):
...
+        for file in files:
+            src_path = os.path.join(root, file)
+            rel_path = os.path.relpath(src_path, source_directory)
+            dst_path = os.path.join(destination_directory, rel_path)
+
+            # Create
+            os.makedirs(os.path.dirname(dst_path), exist_ok=True)
+            os.symlink(src_path, dst_path)
+
+
```
Paul-Ferrell, line 209: I started write that you should mimic how this is done in `builder.py` with `maybe_symlink_copy`, but with some of the options you need to add this way is better.

PR #764, opened 2024-04-15
Preserve extra files permissions

dmageeLANL:
Pavilion wasn't honoring extra files permissions, fixed that by getting the extra file permissions and modifying the file after moving it in _setup_build_dir. This is necessary for executable wrapper scripts which are particularly essential for gpu runs where the target application needs to run on one gpu per process.

Also added the failed return code to the status tracker for failed builds, and fixed the format in the tracker call for failed build moves.

Code review checklist:

### DIFF
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -438,6 +438,8 @@ class TestBuilder:
 
                             try:
                                 self.path.rename(self.fail_path)
+                                tracker.warn("FAILED_BUILD_MOVED to {}"
+                                             .format(self.fail_path))
                             except FileNotFoundError as err:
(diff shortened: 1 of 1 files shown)


Paul-Ferrell:
Rejecting. 
All permissions _except_ write permissions and those masked by your umask should be preserved. 
Pavilion's umask can be set in the pavilion.yaml.
If you want to edit (rather than just overwrite) a file from the build, put it in build.copy_files.
It will get actually copied, and user/group write permissions will be set.
```yaml
test:
  build:
    source_path: my_source.tgz
    extra_files:
      - some_file.txt
    copy_files:
      - some_file.txt
```

Issue #763, opened 2024-04-12
Build fail on node quits badly.


dmageeLANL:
Build on a node and the build fails but the status file looks like:

``` bash
1712858170.457788 STATUS_CREATED Created status file.
1712858170.458920 CREATED Test directory and status file created.
1712858170.637588 BUILD_CREATED Builder created.
1712858170.642490 CREATED Test directory setup complete.
1712858171.290539 SCHEDULED Test kicked off (individually (flex)) under slurm scheduler.
1712858553.553697 BUILD_CREATED Builder created.
1712858904.348879 BUILD_CREATED Builder created.
1712858905.195227 BUILD_WAIT Waiting on lock for build 20796ce600e76a40.
1712858905.195858 BUILDING Starting build 20796ce600e76a40.
1712858905.196529 BUILDING Copying source directory /usr/projects/hpcml/test/dmagee/venado-acceptance/test_src/parthenon for build /usr/projects/hpcml/test/dmagee/venado-acceptance/../working_dir/builds/20796ce600e76a40 as the build directory.
1712858909.576306 BUILDING Generating dynamically created files.
1712858924.474080 BUILD_FAILED Build returned a non-zero result.
1712858925.552891 PREPPING_RUN Converting run template into run script.
1712858925.553972 RUNNING Starting the run script.
1712858925.557292 RUNNING Currently running.
1712858928.135093 RESULTS Parsing 7 result types.
1712858928.138710 RESULTS Performing 2 result evaluations.
1712858928.140112 RESULTS_ERROR Unexpected error parsing results: 'unsupported operand type(s) for /: 'NoneType' and 'int''... (This is a bug, you should report it.) See 'pav log kickoff 34' for the full error.
```

And then the build directory never gets moved to the 'failed' build directory and there's no record of the build anywhere. So if you go into test_runs/{number}/build/build you can't use ccmake to examine the artifacts.  The build.log exists though.

Also, it doesn't stop the other builds. Perhaps there should be a 'fail_all` option for the build (and run) key where, if it's set, one build fail causes all runs to fail and if it isn't, they all act independently so build failures don't stop the series.

I think the solution would be to have it do what it seems it intends to and:
* Move the failed build to the failed build directory.
* Notify the user of the path to the failed build directory.  
* Have the cancel event actually call scancel and kill the other jobs.

But the second part of that is debatable. I think it's better to have the default setting be: don't kill other jobs if one fails, but that should be an option. 

[...]

PR #765, opened 2024-04-16
Fixed on_node failed build non-stopping problem. Now builds with on_n…

dmageeLANL:
Fixes issue #763. Raises exception on non test.build_local build failure which stops pavilion from running the test anyway. And stops non test.build_local runs from writing `BUILD_CREATED` lines to status over and over when this test is queried.

There could probably be a better condition to stop writing BUILD_CREATED statuses. But in this case, if the run is completed or has an error status, it will keep that status.

Code review checklist:

### DIFF
--- a/lib/pavilion/builder.py
+++ b/lib/pavilion/builder.py
@@ -96,7 +96,9 @@ class TestBuilder:
 
         self.path = working_dir/'builds'/self.name  # type: Path
 
-        if not self.path.exists():
+        current_status = status.current()
+
+        if not self.path.exists() and "ERROR" not in current_status.state:
             status.set(state=STATES.BUILD_CREATED, note="Builder created.")
 
         self.tmp_log_path = self.path.with_suffix('.log')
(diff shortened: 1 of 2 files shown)


dmageeLANL:
This PR now has a bonus fix! The original PR revealed a bug. In status_file.py:_parse_status_line, if the status timestamp is in the legacy format it can't be cast to a float and so the program skips into the first except. There it pops the status line list again, and tries to apply the legacy time format to that element. But that element is now the 2nd element which is the status state not the timestamp. So don't double pop, just reuse `time_part`. 

Also, the failsafe call `when = datetime.datetime(0, 0, 0)` doesn't work either. If you don't want it to fail, use `when = datetime.datetime.now()`.

This commit caught that because now `status.current()` is called in `builder.py:TestBuilder:__init__()`. And this is REALLY in the main line of the code. It's is called nearly every time the `pav` command is run. So when it hit the legacy test, it failed.

Paul-Ferrell (approved):

PR #770, opened 2024-06-07
Fix max-digits in output formatting to convert input to int

hwikle-lanl:
Output formatting code took max-digits argument as a string, and never converted it to an int, resulting in an eventual TypeError. This commit simply modifies the argument parser to automatically convert the value into an integer.

Code review checklist:


Paul-Ferrell (commented):
Looks fine.

PR #782, opened 2024-08-14
Add suites directory

### LINKED ISSUE #742: Create 'suites' directory, test re-org
We should add a new config dir - `suites`, and deprecate the test_src and tests directory.

In a suites directory:
- a single yaml file can contain a suite named for that file (as per `tests` currently)
- a directory, containing:

  - a suite.yaml file (required), with the suite configuration.
  - a hosts.yaml file (optional) containing host specific configs (unlike regular hosts files, these will have multiple configs (see below)
  - a modes.yaml file (option) with multiple modes
  - an os.yaml file (option) with OS options
- Pavilion should look in the suite directory first for any source files

  - If a source path should default to the series' directory if no path or download is specified.
  - The yaml files listed above should be ignored when calculating the build hash.

## Host/Mode Files

The host and mode files in a suite directory should have the same format as a test, except the keys should be host/mode/os names.

```yaml
rocinante:
  variables:
    foo: 'bar'

crossroads:
  variables:
    foo: 'baz'
```

hwikle-lanl:
Code review checklist:


Paul-Ferrell:
I'm pretty sure you're not actually ready for me to review this, but I see a few issues already.
The expected directory structure is:
`suites/<suite_name>/`
So for whatever suite you're looking at, you should be finding files under the specific directory for a suite rather than the general `suites` directory.

lib/pavilion/filters/parse_time.py
```diff
@@ -137,3 +142,24 @@ def normalize(unit: str) -> str:
         return unit + "s"
 
     return unit
+
+
+def safe_update(date: datetime,
+                year: int = None,
+                month: int = None, day: int = None) -> datetime:
```
Paul-Ferrell: Docstring needed.

lib/pavilion/filters/parse_time.py
```diff
@@ -62,12 +63,16 @@ def parse_duration(rval: str, now: datetime) -> datetime:
         raise ValueError(f"Invalid unit {unit} for duration")
 
     if unit == 'years':
-        return now.replace(year=now.year - mag)
+        return safe_update(now, year=now.year - mag)
```
Paul-Ferrell, line 67: It feels like using datetime.timedelta would be good enough, but this is better.

lib/pavilion/commands/config.py
```diff
@@ -244,7 +244,8 @@ def create_config_dir(self, pav_cfg: config.PavConfig, path: Path,
             raise ConfigCmdError("Error writing config file at '{}'"
                                  .format(config_file_path), err)
 
-        for subdir in 'hosts', 'modes', 'tests', 'os', 'test_src', 'plugins', 'collections':
+        for subdir in ('hosts', 'modes', 'tests', 'os', 'test_src',
+            'plugins', 'collections', 'suites'):
```
Paul-Ferrell: Since we're replacing test_src and tests with suites, let's get rid of those here.

lib/pavilion/builder.py
```diff
@@ -965,17 +978,33 @@ def _hash_io(cls, contents):
...
-            'hash' based on it's name and mtime, assuming we've run _date_dir
-            on it before hand. This produces an arbitrary string, not a hash.
-        :param Path path: The path to the directory.
-        :returns: The 'hash'
+    @classmethod
+    def _hash_dir(cls, path: Path, exclude: List[str] = None) -> str:
+        """Recursively hash the files in the given directory, excluding
+        files with the given suffixes.
+
+        We might need to modify this to handle circular symlinks.
```
Paul-Ferrell: We definitely need to handle this.

Paul-Ferrell: Or, just don't follow symlinks (simply add their pointer dest to the hash).

lib/pavilion/series/test_set.py
```diff
@@ -722,14 +725,22 @@ def wait(self, wait_for_all=False, wait_period: int = TEST_WAIT_PERIOD) -> int:
...
+        timeout = set_default(timeout, math.inf)
+
         # No tests to wait for
         if not self.started_tests:
             return 0
 
         completed_tests = self.mark_completed()
 
+        start = time.monotonic()
+
```
Paul-Ferrell: Using the monotonic clock really seems like over kill.  I'd like to hear why it matters.

lib/pavilion/resolver/resolver.py
```diff
@@ -636,7 +735,9 @@ def _load_raw_configs(self, request: TestRequest, modes: List[str],
 
             # Apply modes.
             try:
-                test_cfg = self.apply_modes(test_cfg, modes)
+                test_cfg = self.apply_modes(test_cfg, modes, request.suite)
+                test_cfg = self.apply_host(test_cfg, self._host, request.suite)
+                test_cfg = self.apply_os(test_cfg, self._os, request.suite)
```
Paul-Ferrell: The stack should be:
 - modes
 - test_layers
 - host
 - os
 - <empty config>

What you have is very backwards.

Paul-Ferrell: IE settings in `host` should override `os`, and so on.

lib/pavilion/filters/parse_time.py
hwikle-lanl: `timedelta` unfortunately can't handle variable-size units like months and years, which is what makes this necessary.

lib/pavilion/commands/config.py
hwikle-lanl: Since we're technically deprecating those, would it make sense to keep them around for awhile for backwards compatibility? Otherwise we'll have to restructure all the existing tests as soon as this is rolled out.

lib/pavilion/series/test_set.py
hwikle-lanl: The choice was motivated by this note in the docs for `time.time`:

> Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second. While this function normally returns non-decreasing values, it can return a lower value than a previous call if the system clock has been set back between the two calls.

hwikle-lanl:
Merging, since all feedback from @Paul-Ferrell  is implemented.

PR #783, opened 2024-08-23
Run results error

### LINKED ISSUE #781: run_result being accessed before assignment.
We seem to be hitting an issue with result parsing in Pavilion.  Seems like it's trying access run_result before assignment.  Oddly, it's only happening on some tests.  Maybe when the test application has completed output, but not returned a value and the pavilion timeout is hit.  Results in a results.json with only 5 values in it.  The results.log is empty.  Running pav results -L -r <test_id> results in a full log that shows a full test result before adding result parsed entries, it parses the table as instructed, but then fails with the run_result issue and ends up with a very sparse results.json saved to disk.

hwikle-lanl:
Code review checklist:

### DIFF
--- a/lib/pavilion/commands/_run.py
+++ b/lib/pavilion/commands/_run.py
@@ -237,7 +237,7 @@ class _RunCommand(Command):
             return
         except TimeoutError:
             # This is expected
-            pass
+            return
         except Exception:
             # Some other unexpected exception.
             test.status.set(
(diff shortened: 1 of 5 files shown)


lib/pavilion/commands/_run.py
```diff
@@ -234,9 +234,10 @@ def _run(self, test: TestRun):
         except TestRunError as err:
             # An unexpected TestRunError
             test.status.set(STATES.RUN_ERROR, err)
+            return
         except TimeoutError:
             # This is expected
-            pass
+            return
```
francinelapid: is this adding a status to the test like the other exceptions? If not, can you add one?

francinelapid (changes requested):

francinelapid (approved):

PR #785, opened 2024-09-21
The 'errors' record for series now counts all errors for tests/series.

Paul-Ferrell:
I've noticed that the 'errors' number when running `pav series status` doesn't include all test errors, like timeouts and other errors. It should with this change.

Code review checklist:

### DIFF
--- a/lib/pavilion/series/info.py
+++ b/lib/pavilion/series/info.py
@@ -149,21 +149,30 @@ class SeriesInfoBase(Mapping):
 
         status_obj = self._get_status_file()
 
+        error_values = [
+            'ERROR',
+            'CANCELLED',
+            'TIMEOUT',
+            ]
+
         errors = 0
         for status in status_obj.history():
-            if status.state in (status_file.SERIES_STATES.ERROR,
-                                status_file.SERIES_STATES.BUILD_ERROR,
-                                status_file.SERIES_STATES.CREATION_ERROR,
-                                status_file.SERIES_STATES.KICKOFF_ERROR):
-                errors += 1
+            for error_val in error_values:
+                if error_val in status.state:
+                    errors += 1
+                    break
 
         for test_path in self._tests:
             test_info = self.test_info(test_path)
             if test_info is None:
                 continue
 
-            if test_info.result == TestRun.ERROR:
-                errors += 1
+            # Look for any bad test states
+            for error_val in error_values:
+                for state in test_info.state_history:
+                    if error_val in state.state:
+                        errors += 1
+                        break
 
         return errors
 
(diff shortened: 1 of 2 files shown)


lib/pavilion/series/info.py
```diff
@@ -149,21 +149,30 @@ def errors(self) -> int:
...
+
         errors = 0
         for status in status_obj.history():
-            if status.state in (status_file.SERIES_STATES.ERROR,
-                                status_file.SERIES_STATES.BUILD_ERROR,
-                                status_file.SERIES_STATES.CREATION_ERROR,
-                                status_file.SERIES_STATES.KICKOFF_ERROR):
-                errors += 1
+            for error_val in error_values:
+                if error_val in status.state:
```
hwikle-lanl, line 161: Using `str.__contains__` here feels a bit kludgey and error-prone. A better solution would be to implement a mechanism to separate statuses into `ERROR`, `CANCELLED`, and `TIMEOUT` classes, either by:

a) Creating something like a `class` field on `TestStatusInfo` that takes on an `enum` value; or

b) By implementing three static methods that return sets of `ERROR`, `CANCELLED`, and `TIMEOUT` statuses respectively and then checking whether `status.state` is contained within each set (or simply the union of the three sets).

hwikle-lanl, line 161: Using `str.__contains__` here feels a bit kludgey and error-prone. A better solution would be to implement a mechanism to separate statuses into `ERROR`, `CANCELLED`, and `TIMEOUT` classes, either by:

a) Creating something like a `class` field on `TestStatusInfo` that takes on an enum value and testing against that; or

b) Implementing three static methods on `StatesStruct` that return sets of `ERROR`, `CANCELLED`, and `TIMEOUT` statuses respectively and then checking whether status.state is contained within each set (or simply the union of the three sets).

lib/pavilion/series/info.py
```diff
@@ -149,21 +149,30 @@ def errors(self) -> int:
...
             test_info = self.test_info(test_path)
             if test_info is None:
                 continue
 
-            if test_info.result == TestRun.ERROR:
-                errors += 1
+            # Look for any bad test states
+            for error_val in error_values:
+                for state in test_info.state_history:
+                    if error_val in state.state:
```
hwikle-lanl, line 173: Ditto here.

hwikle-lanl (changes requested):

lib/pavilion/series/info.py
hwikle-lanl: The second option feels cleaner to me.

PR #803, opened 2025-01-24
Handle filesystem latency when creating `RUN_COMPLETE` file

### LINKED ISSUE #802: Uncaught exception when creation of `RUN_COMPLETE` file times out
Need to add logic to handle this gracefully.

hwikle-lanl:
Code review checklist:

### DIFF
--- a/lib/pavilion/test_run/test_run.py
+++ b/lib/pavilion/test_run/test_run.py
@@ -832,10 +832,11 @@ class TestRun(TestAttributes):
 
         return ret
 
-    def _create_complete_file(self, complete_tmp_path: Path) -> None:
+    @staticmethod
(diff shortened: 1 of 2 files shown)


Paul-Ferrell (commented):
I really don't like throwing an exception here. 
I think we should try harder to force a cache update. Instead of just doing an `exists`, you might try tossing in a `path.parent.iterdir()`. Apparently opening the parent directory should cause the directory cache to refresh.

If that doesn't seem to be working, drop a symlink to the temp file location where you were going to move the original file, and walk away instead of throwing an exception.

Also, CREATION_ERROR is explicitly for test creation problems. I'd create a new 'WARNING' label for cases like this.

smehta99 (approved):
LGTM

PR #813, opened 2025-02-05
Prevent tests from transitioning to BUILD_CREATED state after failure

### LINKED ISSUE #804: Test state transitions to `BUILD_CREATED` after build failure.
`pav status` will sometimes show a `BUILD_CREATED` state even after a test fails to build. In addition, when this occurs, there are multiple entries for `BUILD_CREATED` in the status log.

hwikle-lanl:
Previously, tests were failing during the build phase and correctly logging a `BUILD_ERROR` state, but then incorrectly continuing on to log a `BUILD_COMPLETE` state, sometimes multiple times. This commit fixes that. It also fixes a double logging of an ABORTED state (albeit with different messages).

Several issues with logging states remain:

- In the aforementioned double ABORTED state, one log statement should probably be removed.

- Nearly identical tests will show either an ENV_FAILED state, or an ENV_FAILED state followed by a BUILD_FAILED state. This inconsistency needs to be resolved, but a decision needs to be made as to the intended behavior.

- Some log states appear to be used inappropriately, which should eventually be fixed.

Code review checklist:

### DIFF
--- a/lib/pavilion/status_file.py
+++ b/lib/pavilion/status_file.py
@@ -71,6 +71,7 @@ Known States:
         """Validate all of the constants."""
 
         self._help = {}
+        self._fatal_states = {self.STATUS_ERROR}
 
         # Validate the built-in states
         for key in dir(self):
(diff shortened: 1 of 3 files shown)


lib/pavilion/status_file.py
```diff
@@ -109,6 +110,11 @@ def list(self):
         """List all the known state names."""
         return self._help.keys()
 
+    def is_fatal(self, state) -> bool:
+        """Determine whether the given state is fatal."""
+
+        return state in self._fatal_states
+
```
Paul-Ferrell, line 117: Yes please.

Paul-Ferrell (approved):

PR #827, opened 2025-03-26
Allow for setup in existing directories.

Paul-Ferrell:
Code review checklist:


lib/pavilion/commands/config.py
```diff
@@ -205,7 +205,7 @@ def create_config_dir(self, pav_cfg: config.PavConfig, path: Path,
 
         # This should fail if it already exists.
```
hwikle-lanl: Remove this comment for consistency.

hwikle-lanl (changes requested):

lib/pavilion/commands/config.py
Paul-Ferrell: I realized that the subdirs should also get an `exist_ok=True`.

hwikle-lanl (approved):

PR #834, opened 2025-04-02
817 automatically perform module purge before building tests

### LINKED ISSUE #817: Automatically perform `module purge` before building tests

hwikle-lanl:
Code review checklist:

### DIFF
--- a/lib/pavilion/commands/run.py
+++ b/lib/pavilion/commands/run.py
@@ -108,6 +108,11 @@ class RunCommand(Command):
                  "under which Pavilion runs has changed."
         )
         parser.add_argument(
+            '--no-purge', action='store_false', default=True, dest='purge',
+            help="Do not perform a module purge before running tests. Modules "
+                 "are purged by default."
+        )
+        parser.add_argument(
             'tests', nargs='*', action='store', metavar='TEST_NAME',
             help='The name of the tests to run. These may be suite names (in '
                  'which case every test in the suite is run), or a '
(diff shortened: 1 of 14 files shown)


Paul-Ferrell:
This should be an option to build/run, but the default should be 'true'.

Paul-Ferrell:
I saw that you added a commit that module purges only when modules are specified.  It should be always (if enabled), with checks for whether a module system is present (IE - does the `module` alias exist?

lib/pavilion/commands/run.py
```diff
@@ -107,6 +107,11 @@ def _generic_arguments(parser):
                  "should be necessary only if the system or user environment "
                  "under which Pavilion runs has changed."
         )
+        parser.add_argument(
+            '--no-purge', action='store_false', default=True, dest='purge',
+            help="Do not perform a module purge before running tests. Modules "
+                 "are purged by default."
+        )
```
Paul-Ferrell: I would expect this option as a test config option, not as a command line argument. I'd expect it to be set at the host file level.

lib/pavilion/scriptcomposer.py
```diff
@@ -129,6 +129,11 @@ def module_change(self, module, sys_vars, config_wrappers):
 
         self.env_change(mod_env)
 
+    def module_purge(self) -> None:
+        """Add a module purge to the script."""
+
+        self._script_lines.append("module purge")
+
```
Paul-Ferrell: This will throw an error if the `module` command doesn't exist.

Paul-Ferrell (changes requested):

Paul-Ferrell (approved):

PR #882, opened 2025-07-11
Pflarr/fix autoexit

Paul-Ferrell:
Code review checklist:


Paul-Ferrell:
This addresses an issue I found when doing PE testing this morning. 
If you set `autoexit` to false, there's now a `echo` statement between the last command and the script end that will cause the test to pass even if it should have failed.
The other issues was that the `err` bash function should do an `exit 1` rather than a `return 1`, so that script failure is forced even when `autoexit` is false.

PR #898, opened 2025-11-25
Fix missing test configs

### LINKED ISSUE #894: Config files missing when running Pavilion
I sometimes run into issues, e.g. when running the post-DST testing, where one or two of the runs of a series do not get a config file created (or it is deleted shortly after creation), so doing `pav status` on that series results in the error "Could not find config file for test". Checking the run dir for that test shows that there is, indeed, no config file.

I also just ran into an issue on Venado where I submitted some runs with Pavilion (the OSU suite) on Friday and `pav status` worked fine (i.e. gave the expected output) on Friday, but today (Wed.) all the config files for those runs are gone and `pav status` gives me the corresponding error "Could not find config file for test". It's test series s719-s724 in case that is helpful. I'm mentioning this because I suspect that it is related.

I don't know how to even start debugging this so I don't have any more information than this, sorry! I am using tmux in case that could be a factor.

hwikle-lanl:
Code review checklist:

### DIFF
--- a/lib/pavilion/test_run/test_run.py
+++ b/lib/pavilion/test_run/test_run.py
@@ -591,18 +591,23 @@ class TestRun(TestAttributes):
 
         try:
             wait(
-                lambda: tmp_path.exists(),
+                tmp_path.exists,
                 interval=0.1,
                 timeout=5,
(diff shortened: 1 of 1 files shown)


cielling (approved):
approve

PR #905, opened 2025-12-11
Update _kickoff docstring with across_nodes information

### LINKED ISSUE #861: Documentation is missing `across_nodes`
It looks like the documentation for the `across_nodes` scheduler option is missing from the ReadTheDocs for Pavilion2 as well as from the `pav show config` output.
Here is the RTD page that looks like it should contain the documentation for `across_nodes`: https://pavilion2.readthedocs.io/en/latest/source/schedulers.html#pavilion.schedulers.SchedulerPlugin._kickoff

hwikle-lanl:
Code review checklist:

### DIFF
--- a/docs/tests/scheduling.rst
+++ b/docs/tests/scheduling.rst
@@ -23,7 +23,7 @@ Pavilion comes with three scheduler plugins:
     -----------+------------------------------------------------------
      raw       | Schedules tests as local processes.
      slurm     | Schedules tests via the Slurm scheduler.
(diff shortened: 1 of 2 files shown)


tygoetsch (changes requested):
Looks good. I think there's a decent distinction between between what across_nodes and include_nodes does with this change. The only thing would be a command-line example along-side the yaml example of the syntax used.

tygoetsch (approved):
Yep, that looks good to me!

PR #928, opened 2026-03-06
Add PBS scheduler plugin

### LINKED ISSUE #922: PBS Scheduler Support
A former coworker wrote a PBS scheduler plugin for Pavilion that worked quite well. Unfortunately, that employee has moved on and the version of Pavilion at my former workplace was outdated when I got there 4 years ago. We (and myself, in particular) love this application and want to keep using it. However to implement it, it will need to work with PBS at this new position. Would it be possible to request assistance with reviewing this PBS plugin to see how feasible it might be to get it working with the devel branch of Pavilion2 or at least give me some pointers in the right direction?

I am a fairly competent Python programmer (learning more and more every day) and I have reviewed it myself along with the API documentation, but it is a bit too deep for me with my current experience and knowledge level of the API just using the Pav documentation.

Thank you so much!

[pbs.py.txt](https://github.com/user-attachments/files/25295573/pbs.py.txt)

[yapsy-plugin.txt](https://github.com/user-attachments/files/25295602/yapsy-plugin.txt)

hwikle-lanl:
Code review checklist:

### DIFF
--- a/lib/pavilion/schedulers/plugins/pbs.py
+++ b/lib/pavilion/schedulers/plugins/pbs.py
@@ -72,6 +72,13 @@ def validate_pbs_states(states: List[str]) -> List[str]:
 class PBSVars(SchedulerVariables):
     """Scheduler variables for the Pbs scheduler."""
 
+    EXAMPLE = SchedulerVariables.EXAMPLE.copy()
+    EXAMPLE.update({
+        "test_cmd": "mpirun -np 16 --host node01:8,node02:8",
+        "walltime": "00:01:00",
+        "queue": "normal"
+    })
+
     def _test_cmd(self) -> str:
         """Construct a cmd to run a process under this scheduler, with the
         criteria specified by this test.
@@ -107,12 +114,16 @@ class PBSVars(SchedulerVariables):
 
     @dfr_var_method
     def walltime(self):
+        """Walltime to add to job in PBS format."""
+
         walltime = self._sched_config["pbs"]["walltime"]
 
         return walltime
 
     @dfr_var_method
     def queue(self):
+        """Queue to run job in."""
+
         if self._sched_config["pbs"]["queue"] is not None:
             return self._sched_config["pbs"]["queue"]
         else:
--- a/lib/pavilion/schedulers/plugins/pbs.py
+++ b/lib/pavilion/schedulers/plugins/pbs.py
@@ -408,9 +408,6 @@ class PBS(SchedulerPluginAdvanced):
         _ = self
         cmd = ["qsub"]
 
-        print("Scheduler variables:")
-        print(json.dumps(sched_config["pbs"], indent=5))
-
         if sched_config["pbs"]["walltime"]:
             cmd.append("-l walltime={}".format(sched_config["pbs"]["walltime"]))
 
--- a/lib/pavilion/schedulers/plugins/pbs.py
+++ b/lib/pavilion/schedulers/plugins/pbs.py
@@ -7,11 +7,12 @@ import shutil
 import subprocess
 import time
 import json
-from typing import List, Union, Any, Tuple, Dict
+from typing import List, Union, Any, Tuple, Dict, Optional
 
 import hostlist
 import yaml_config as yc
 from pavilion import sys_vars
+from pavilion.config import PavConfig
 from pavilion.jobs import Job, JobInfo
 from pavilion.status_file import STATES, TestStatusInfo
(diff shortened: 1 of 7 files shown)


lib/pavilion/schedulers/plugins/pbs.py
```diff
@@ -0,0 +1,648 @@
...
+        queue = self._config["pbs"]["queue"]
+        if queue is not None:
+            lines.append("#PBS -q {}".format(queue))
+
+        # Mandantory Account name
+        if self._sched_vars.account() != "":
+            lines.append("#PBS -A {}".format(self._sched_vars.account()))
+
+        if self._sched_vars.walltime() != "":
+            walltime = "00:01:00"
```
hwikle-lanl: This will set the walltime to "00:01:00" regardless of what's in the config.

lib/pavilion/schedulers/plugins/pbs.py
```diff
@@ -0,0 +1,648 @@
...
+
+        if tasks is None:
+            tasks = int(self.tasks_per_node()) * nodes
+
+        if self._sched_config["pbs"]["mpi_cmd"] != "":
+            cmd = ["mpirun"]
+            cmd.extend(self.mpirun_opts())
+            cmd.extend(["--host", ",".join(self._nodes.keys())])
+
+        return " ".join(cmd)
```
hwikle-lanl: This will raise a `RuntimeError` when `mpi_cmd` is set to the empty string, since `cmd` won't be defined.

lib/pavilion/schedulers/plugins/pbs.py
```diff
@@ -0,0 +1,648 @@
...
+    @dfr_var_method
+    def walltime(self) -> str:
+        """Walltime to add to job in PBS format."""
+
+        walltime = self._sched_config["pbs"]["walltime"]
+
+        return walltime
+
+    @dfr_var_method
+    def queue(self) -> str:
```
hwikle-lanl: Return type should be `Optional[str]`

lib/pavilion/schedulers/plugins/pbs.py
```diff
@@ -0,0 +1,648 @@
...
+            "qsub_extra": validate_list,
+            "walltime": pbs_str,
+            "queue": pbs_str,
+            "mpi_cmd": validate_mpi,
+        }
+
+        return elems, validators, defaults
+
+    @classmethod
+    def parse_node_list(cls, node_list: List[str]) -> NodeList:
```
hwikle-lanl: `node_list` type should be `Union[str, List[str], None]`

lib/pavilion/schedulers/plugins/pbs.py
```diff
@@ -0,0 +1,648 @@
...
+        return NodeList(nodes)
+
+    def _get_alloc_nodes(self, job: Job) -> NodeList:
+        """Get the list of allocated nodes."""
+
+        _ = job
+        node_list = []
+
+        nodefile = os.environ["PBS_NODEFILE"]
+        with open(nodefile, "r") as fin:
```
hwikle-lanl: This needs a `try..except` block in case `PBS_NODEFILE` doesn't exist, can't be opened, etc.

lib/pavilion/schedulers/plugins/pbs.py
```diff
@@ -0,0 +1,648 @@
...
+    def _filter_custom(self,
+                       sched_config: Dict[str, Any],
+                       node_name: str,
+                       node: NodeInfo) -> Optional[str]:
+        """Filter nodes by features. (Returns why a node should be filtered out, or None if it
+        shouldn't be."""
+
+        _ = self
+
+        pbs_config = sched_config["pbs"]
```
hwikle-lanl: This is dead code. If it doesn't make sense to define this method for PBS, just return `None`.

lib/pavilion/schedulers/plugins/pbs.py
```diff
@@ -0,0 +1,648 @@
...
+
+        return results
+
+    # Pbs status mappings
+    # SCHED_WAITING - The job is still queued and waiting to start.
+    SCHED_WAITING = [
+        "Q",
+        "W",
+    ]
+    # SCHED_OK - From pavilion's perspective, these all mean Pavilion should
```
hwikle-lanl: Comment doesn't match variable name.

hwikle-lanl (commented):
Also should add some unit tests and update docs.

PR #929, opened 2026-03-09
Avoid Division by Zero in Outliers Expression Function

hwikle-lanl:
This pull request handles two special cases in the outliers expression function which result in division by zero:

1. The case where an empty list is passed to the function.
2. The case where the standard deviation of the values is zero (i.e. all values in the list are the same).

In both cases, the function now returns an empty outliers dictionary, which is a reasonable behavior in both instances.

Code review checklist:


cielling (approved):
Looks good!

PR #933, opened 2026-03-09
Fix `command` Result Parser

hwikle-lanl:
Code review checklist:


Paul-Ferrell (changes requested):
You don't need to refactor anywhere near this much:

```python
             if not rconf["working_dir"].startswith("/"):
                rconf['working_dir'] = test.path / BUILD_DIR / rconf["working_dir"]
            else:
                rconf['working_dir'] = Path(rconf["working_dir"])
```

hwikle-lanl:
Clarification from @Paul-Ferrell : Rather than having every result parser take `working_dir` as an argument to its `__call__` method, use `inspect` to only pass the working directory to the result parser if the `__call__` method has an argument named `working_dir`.

Paul-Ferrell (approved):

PR #1046, opened 2026-04-09
Fix tests placed in wrong working directory when multiple configs are used

### LINKED ISSUE #1045: Test runs placed in wrong working directory when multiple config directories are used

hwikle-lanl:
These changes fix an issue in which test runs were sometimes placed in the wrong working directory when multiple config directories were being used.

In addition, these changes:
- Add unit tests for multi-config use cases.
- Reorganize unit tests such that each unit test suite has its own Pavilion configuration directory and working directory, done to support the new tests, but which should also make diagnosing unit test failures easier.
- Fix an issue, discovered during unit testing, that causes the wrong suite to sometimes be loaded if suites of the same name exist in different configuration directories.


cielling (approved):
This looks good to me.

PR #1052, opened 2026-04-16
Fix spack TTY error and orphaned processes

hwikle-lanl:
These changes fix a pair of bugs:

1. The `build` and `run` methods would leave behind orphaned processes when killing subprocesses, if those subprocesses had themselves started their own subprocesses.

2. When building with Spack, Spack would issue a traceback complaining that it could not perform TTY-related operations.


smehta99 (approved):
LGMT

PR #1072, opened 2026-05-12
fix(log): handle invalid test/series IDs without uncaught ValueError

### LINKED ISSUE #1064: Uncaught `ValueError` when passing bad test IDs to `pav log`

SAY-5:
Code review checklist:

- [x] Code is generally sensical and well commented
- [x] Variable/function names all telegraph their purpose and contents
- [x] Functions/classes have useful doc strings
- [x] Function arguments are typed
- [x] Added/modified unit tests to cover changes.

Fixes #1064.

When `pav log <subcmd> <id>` is passed an id that is neither a valid TestID nor a valid SeriesID, `LogCommand.run` falls through to `SeriesID(args.id)`, whose constructor raises an uncaught `ValueError` and crashes the CLI.

This guards the conversion with `SeriesID.is_valid_id` and prints a readable error before returning 1, matching how the other "load failed" paths in this command already behave. Adds a regression test.


SAY-5:
Retargeted this to stable and rebased the branch onto it, so the PR is now just the single log.py guard plus its test.

PR #1076, opened 2026-06-09
Fix `null` result in results log

hwikle-lanl:
This commit fixes a race condition in which result loggers sometimes read `results.json` before it has been fully written, resulting in `null` results and missing keys in the result log.


cielling (approved):
I think this looks good.

